PR #7, opened 2021-04-19
Adds beachball to start collecting change files

kenotron:
This adds a collection of things to enable `beachball` workflow:

1. adds a PR & Release github actions
2. adds a `yarn change` command for manually creating change files
3. adds a check in those github actions to gate any changes without change files

PATs are required for the actual releases (we're waiting for a service account to be created), so therefore the release.yml is manually triggered for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,9 @@
     "test": "lage test",
     "lint": "lage lint",
     "lage": "lage",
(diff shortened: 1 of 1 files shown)


alloy:
Looks simple enough 👌

PR #8, opened 2021-04-20
Fixes various publishConfig's to allow for publishing to npmjs; adds a package-template folder

kenotron:
This PR USED to be about checking the gates, now it will do the following:

1. removes foo & bar packages
2. moved bar code to /package-templates (not affecting the yarn workspace)
3. adds publishConfig access: public


kenotron:
After this is in, we can try manually triggering the release workflow

alloy:
Nice, let’s do it 🕺

PR #9, opened 2021-04-20
auto release on merge of main

kenotron:
This will enable the release to occur every time the main is merged. The release job will only actually release things when change files are detected. Otherwise, it will simply run a build & test on the main branch at that time.


alloy:
Aw yeah 🚀

PR #14, opened 2021-04-23
Publishing cleanup

alloy:


change/monorepo-scripts-3c5e63c1-709c-4d5d-af40-308b1d3df4ce.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Literally none",
+  "packageName": "monorepo-scripts",
+  "email": "eloy.de.enige@gmail.com",
+  "dependentChangeType": "none"
+}
```
alloy, line 7: @kenotron It keeps thinking I made changes to this, any idea why?

kenotron: Flip that project to "private: true" perhaps would make it stop :)

scripts/just.config.ts
```diff
@@ -6,7 +6,7 @@ export const types = tscTask({ emitDeclarationOnly: true });
 
 export const build = () =>
   esbuildTask({
-    entryPoints: glob.sync("src/**/*.ts"),
+    entryPoints: glob.sync("src/**/*.{ts,tsx}"),
```
kenotron, line 9: you technically did make changes in /scripts which has a package.json in the "workspace" - flip that package.json to have private flag would stop that noise.

package.json
```diff
@@ -14,6 +14,7 @@
     "test": "lage test",
     "lint": "lage lint",
     "lage": "lage",
+    "ci": "yarn build && yarn test && yarn lint && yarn checkchange",
```
kenotron, line 17: nice - I think all you need to do this: `yarn lage build test lint && yarn checkchange`

kenotron (commented):
This is good enough as-is, but I have some good suggestions that you should probably consider.

scripts/just.config.ts
alloy: Yes, very true, my bad. I feel like I keep having to do it for this package but raised it at the wrong time. Earlier on there was a mode change from 644 to 755 iirc.

package.json
alloy: :boom: Went from 4 sec to 3 sec

scripts/just.config.ts
alloy: Added `private: true` 👍

PR #19, opened 2021-04-23
bumps tools so they boot up faster

kenotron:
I have identified some source of slowness in the tools. This bumps those deps to make the dev experience way better :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,8 +21,8 @@
     "release": "yarn beachball publish"
   },
   "devDependencies": {
-    "beachball": "^2.1.0",
-    "lage": "0.29.0",
+    "beachball": "2.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -21,8 +21,8 @@
     "release": "yarn beachball publish"
   },
   "devDependencies": {
-    "beachball": "^2.1.0",
-    "lage": "0.29.0",
+    "beachball": "2.2.0",
+    "lage": "0.29.1",
```
alloy, line 25: Did you use fixed version requirements here on purpose?

kenotron: Just precautions, I know yarn.lock works - but usually when I upgrade, I'd like to exact on these two tools.

PR #20, opened 2021-04-24
Build cjs and mjs

alloy:


scripts/just.config.ts
```diff
@@ -1,14 +1,34 @@
...
+      ...baseEsbuildOptions,
+      format: "esm",
+      outExtension: { ".js": ".mjs" },
+    }),
+    esbuildTask({
+      ...baseEsbuildOptions,
+      format: "cjs",
+    })
+  );
+};
```
alloy, line 31: @kenotron Am I doing this right?

packages/relay-compiler-language-graphitation/package.json
```diff
@@ -20,7 +20,7 @@
     "monorepo-scripts": "*"
   },
   "publishConfig": {
-    "main": "./lib/index.js",
+    "main": "./lib/index",
```
alloy, line 23: Let module lower pick which format to use

PR #24, opened 2021-05-04
Support explicit concrete type selection

alloy:
In case of an abstract type the user can specify a mock that returns a specific typename, in which case the appropriate fragment spread should be selected.

### DIFF
--- a/packages/graphql-js-operation-payload-generator/src/__tests__/RelayMockPayloadGenerator.test.ts
+++ b/packages/graphql-js-operation-payload-generator/src/__tests__/RelayMockPayloadGenerator.test.ts
@@ -290,10 +290,11 @@ xtest("generate mock using custom mock functions", () => {
   );
 });
 
-xtest("generate mock using custom mock functions for object type", () => {
-  graphql`
+test("generate mock using custom mock functions for object type", () => {
+  const fragment = graphql`
     fragment RelayMockPayloadGeneratorTest8Fragment on Page {
       actor {
+        __typename
         id
         name
       }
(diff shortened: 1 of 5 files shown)


packages/graphql-js-operation-payload-generator/src/__tests__/RelayMockPayloadGenerator.test.ts
```diff
@@ -921,11 +934,12 @@ describe("with @relay_test_operation", () => {
     );
   });
 
-  xtest("generate mock with Mock Resolvers for Interface Type with multiple fragment spreads", () => {
+  test("generate mock with Mock Resolvers for Interface Type with multiple fragment spreads", () => {
```
alloy, line 937: This test best illustrates the change. Previously the type condition of the first fragment that was found was used, in this case `User`. Now the `Page` fragment is selected instead.

packages/graphql-js-operation-payload-generator/src/index.ts
```diff
@@ -227,16 +294,28 @@ function visitDocumentDefinitionNode(
  */
 function reduceToSingleObjectTypeSelection(
   selectionSetNode: SelectionSetNode,
+  explicitTypename: string | undefined,
```
alloy, line 297: This is where the right fragment is specified.

packages/graphql-js-operation-payload-generator/src/index.ts
```diff
@@ -227,16 +294,28 @@ function visitDocumentDefinitionNode(
...
   schema: GraphQLSchema,
   allDocumentDefinitionNodes: ReadonlyArray<DefinitionNode>,
   typeInfo: TypeInfo
 ): SelectionSetNode {
   let reduceToObjectType: GraphQLObjectType | null = null;
```
freiksenet, line 302: I'm confused by this code in general, why do we need to keep it as global state? I won't block merge cause it's been there before, but this is weird.

freiksenet (approved):

packages/graphql-js-operation-payload-generator/src/index.ts
alloy: Unsure which global state you’re referring to, can you elaborate?

PR #25, opened 2021-05-05
Support explicit concrete type selection across mutliple fragment spreads

alloy:
The existing test case didn’t cover what we actually do in TMP.


alloy:
Going to self-merge this and finish up my TMP PR 👍

PR #34, opened 2021-07-02
Add "useMutation" hook

freiksenet:
```ts

 const mutation = graphql`
 mutation SomeReactComponentMutation($newName: String!) {
   someMutation(newName: $newName) {
     __typeName
     id
     newName
   }
 }
 `

 export const SomeReactComponent: React.FunctionComponent<SomeReactComponentProps> = props => {
   const [commitMutation, isInFlight] = useMutation(mutation);
   return (
     <div>
       <button onClick={() => commitMutation({
         variables: {
            newName: "foo"
         },
         optimisticResponse: {
            someMutation: {
              __typename: "SomeMutationPayload",
              id: "1",
              newName: "foo",
            }
         }
       })} disabled={isInFlight}/>
     </div>
   );
 }```

### DIFF
--- a/packages/apollo-react-relay-duct-tape/src/hooks.ts
+++ b/packages/apollo-react-relay-duct-tape/src/hooks.ts
@@ -4,6 +4,7 @@ import invariant from "invariant";
 import {
   useSubscription as useApolloSubscription,
   useQuery as useApolloQuery,
+  useMutation as useApolloMutation,
 } from "@apollo/client";
 
 // import { GraphQLTaggedNode } from "./taggedNode";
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/apollo-react-relay-duct-tape/src/hooks.ts [hunk 1]
```diff
@@ -175,3 +176,86 @@ export function useSubscription<TSubscriptionPayload extends OperationType>(
...
+ *
+ * Returned function can be called to perform the actual mutation with provided variables.
+ *
+ * @param mutation Mutation document
+ * @returns [commitMutationFn, isInFlight]
+ *
+ * commitMutationFn
+ * @param options.variables map of variables to pass to mutation
+ * @param options.optimisticResponse proposed response to apply to the store while mutation is in flight
+ * @returns A Promise to an object with either errors or/and the result data
```
alloy, line 200: Not entirely according to spec, seeing as it’s a disposable in react-relay, but it also doesn’t seem like a stretch to make it a Promise there in the future too https://github.com/facebook/relay/issues/1822, so I’m ok with merging as-is.

alloy (approved):

PR #43, opened 2021-09-29
Make variables optional

freiksenet:
Make variables param optional in mutations to make it simpler to type.


alloy (changes requested):
We can't do this, because it makes it impossible to have TS verify required GraphQL variables.

I.e. consider the following GraphQL operation variable declaration `id: ID!`. Using this optional typing it would become: `{ variables?: { id: string }}` rather than `{ variables: { id: string }}`.

PR #50, opened 2021-10-05
Apollo fragment-based store observing

alloy:
TODO:

- [x] Add pagination to example app
- [x] Pagination fragment requires `@refetchable` directive, is this normally enforced in the compiler or at runtime?
- [x] Introduce better way for ts import transform to know name of query artefact to import
- [x] Split hooks into files
- [x] Compiler needs to extract default variable values and use it in the fragments context
- [ ] Apollo Client watchQuery seems to always yield twice when data updates
- [x] Import refetch/pagination/subscription docs
- [ ] Consider switching from `node` root field to a private reserved field as to not cause any conflicts when reading cached `node` data etc. E.g. `__node` ?
- [x] Consider merging the relay-compiler-language-graphitation package into the apollo-react-relay-duct-tape package, seeing as most-all of it is in service of that package
- [ ] Workout strategy for dev vs prod builds, specifically around stripping out invariant calls
- [ ] Write document on all this:
  * Why?
  * How does the architecture work?
    * The Global Object Identification spec
    * The compiler basics
    * The Node transform
    * The compiler output
    * The hooks and execution vs watch queries
    * The `node` cache field policy
  * How to use it?


examples/supermassive-todomvc/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "ts-node": {
     "transpileOnly": true,
     "compilerOptions": {
-      "module": "ESNext",
+      "module": "commonjs",
```
alloy, line 6: @freiksenet Needed to make this change, which makes sense to me as this is all about running web pack in Node and Node doesn't properly support es6 modules. Am I missing something?

freiksenet: You should pull latest version, it has a fix that works.

freiksenet:
```  "ts-node": {
    "transpileOnly": true,
    "compilerOptions": {
      "module": "ESNext",
      "target": "ES5",
      "moduleResolution": "Node"
    },
    "moduleType": {
      "webpack.config.ts": "cjs"
    }
  },```

freiksenet (approved):

PR #55, opened 2021-10-26
Implement useRefetchableFragment for Apollo Client

alloy:
https://relay.dev/docs/api-reference/use-refetchable-fragment/#internaldocs-banner

(See [this commit](https://github.com/microsoft/graphitation/pull/50/commits/c584e85c60d4b1ffbbe8c5e282729e6a49845d0f) for the compiler artefacts, which I had already previously merged before I decided to PR iterations.)

This currently only works with compiled artefacts, not sure yet if it makes sense in terms of ROI to ever try without compilation.

Does not return the result data of the refetch from the hook directly, instead it returns data by leveraging the watchQuery started from the `useCompiledFragment` hook.


examples/apollo-watch-fragments/src/Todo.tsx
```diff
@@ -1,24 +1,27 @@
...
+import { TodoRefetchQuery } from "./__generated__/TodoRefetchQuery.graphql";
 
 export const Todo_todoFragment = graphql`
-  fragment Todo_todoFragment on Todo {
+  fragment Todo_todoFragment on Todo
+  @refetchable(queryName: "TodoRefetchQuery") {
     id
     description
     isCompleted
+    someOtherField @include(if: $includeSomeOtherField)
```
alloy, line 19: This is truly a horribly uninspired example. Need to come up with a better one that makes sense in the context of this todo app, but for now it at least demonstrates its working.

examples/apollo-watch-fragments/src/Todo.tsx
```diff
@@ -1,24 +1,27 @@
 import React, { useCallback } from "react";
 import {
-  useFragment,
+  useRefetachbleFragment,
```
kerrynf, line 3: Typo in refetch?

Should be useRefetchableFragment I think?

packages/apollo-react-relay-duct-tape/src/hooks.ts
```diff
@@ -148,6 +149,26 @@ export function useFragment<TKey extends KeyType>(
   }
 }
 
+export function useRefetachbleFragment<
```
kerrynf, line 152: Original source of typo I think :)

alloy: 🙀 

Thanks!

PR #59, opened 2021-11-24
[eslint-rules] Hotfix key fields node fixed

vejrj:


alloy:
I understand this is a hot-fix, so I’m ok merging it, but please follow-up with a PR that adds test coverage for this bug.

PR #61, opened 2021-11-29
Suggestion on how to do ESM builds in monorepo

freiksenet:
Currently it's not actually possible to use .mjs generated modules in this package repos - they will end up importing normal js. This is a hacky proposed solution to this, it annotates files with a fully qualified export.


freiksenet:
I need to fix and test all other packages in the monorepo before this can be merged.

scripts/just.config.ts
```diff
@@ -19,15 +40,44 @@ export const build = () => {
...
+        ...baseEsbuildOptions,
+        format: "esm",
+        outdir: "lib/mjs",
+        bundle: true,
+        plugins: [
+          {
+            name: "add-mjs",
+            setup(build) {
+              build.onResolve({ filter: /.*/ }, (args) => {
+                if (args.importer) {
```
alloy, line 54: What's `args.importer`?

freiksenet: This plugin runs whenever esbuild is trying to "bundle" the package. Importer checks that something is importing this package. We rewrite it to include '.js' extension then.

alloy: Gotcha, ta

packages/relay-compiler-language-graphitation/tsconfig.json
```diff
@@ -1,15 +1,13 @@
 {
   "compilerOptions": {
-    "composite": false,
+    "composite": true,
```
alloy, line 3: Hmm, don't recall why I changed this but I think there was a reason for it. I assume at least everything builds, so let's see if I run into an issue again.

freiksenet:
@alloy No need, so I'll move it outside.

examples/supermassive-todomvc/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "tabWidth": 2,
+  "useTabs": false
+}
```
alloy, line 4: No need to block this PR on, but this can just go into the root of the repo?

examples/supermassive-todomvc/src/schema/resolvers.ts
```diff
@@ -0,0 +1,173 @@
+import { v4 } from "uuid";
+// import { Resolvers } from "@graphitation/supermassive";
```
alloy, line 2: Still needed?

package.json
```diff
@@ -9,7 +9,8 @@
   },
   "scripts": {
     "clean": "git clean -fdx -e node_modules",
-    "build": "lage build",
+    "build-only": "lage build",
+    "build": "lage build && lage types",
```
alloy, line 13: 👍

alloy (approved):
:shipit:

PR #62, opened 2021-12-08
[supermassive/doc] Expand on what/why/how

alloy:
Rendered version: https://github.com/microsoft/graphitation/blob/alloy/supermassive-readme/packages/supermassive/README.md

### DIFF
--- a/packages/supermassive/README.md
+++ b/packages/supermassive/README.md
@@ -29,6 +29,83 @@ Similarly, the GraphQL operations themselves, described using e.g. [GraphQL SDL]
 
 In this initial phase, we will achieve the goal of tree-shaking the schema definitions. We do this by inlining required metadata into the documents that describe the operations, after which they can be executed with the need of the entire schema.
 
+Consider a schema like the following:
+
+```graphql
+type Query {
+  me: User!
+}
+
+type User {
+  name: String!
+  presence: Presence!
+}
+
+type Presence {
+  availability: PresenceAvailability!
+}
+
+enum PresenceAvailability {
+  AVAILABLE
+  BUSY
+  OFFLINE
+}
+```
+
+...and a field-resolver map like the following:
+
+```ts
+import { getUser } from "user-service";
+import { getUserPresence } from "presence-service";
+
+const resolvers = {
+  Query: {
+    me: async (_source, _args, context) => getUser(context.currentUserId),
+  },
+  User: {
+    name: (source) => source.name,
+    presence: async (source) => getUserPresence(source.id),
+  },
+};
+```
+
+When given an operation like the following:
+
+```graphql
+query {
+  me {
+    name
+  }
+}
+```
+
+...the compiled output will [conceptually] contain only the following schema definitions:
+
+```graphql
+type Query {
+  me: User!
+}
+
+type User {
+  name: String!
+}
+```
+
+...and the following field-resolver code:
+
+```ts
+import { getUser } from "user-service";
+
+const resolvers = {
+  Query: {
+    me: async (_source, _args, context) => getUser(context.currentUserId),
+  },
+  User: {
+    name: (source) => source.name,
+  },
+};
+```
+
 ### Phase 2
 
 In this phase, we will expand on the previous phase by ahead-of-time compiling the resolution of the operations, their field-resolvers, and invocation thereof into JavaScript code. This essentially does away with any need for AST of the operation during execution.
--- a/packages/supermassive/README.md
+++ b/packages/supermassive/README.md
@@ -27,7 +27,7 @@ Similarly, the GraphQL operations themselves, described using e.g. [GraphQL SDL]
 
 ### Phase 1
 
-In this initial phase, we will achieve the goal of tree-shaking the schema definitions. We do this by inlining required metadata into the documents that describe the operations, after which they can be executed with the need of the entire schema.
+In this initial phase, we will achieve the goal of tree-shaking the schema definitions. We do this by inlining required metadata into the documents that describe the operations, after which they can be executed with the need of the entire schema. This means overall bundle size will be decreased when only a subset of the schema is actually used, which pays off significantly when a host application introduces its first component(s) leveraging GraphQL.
 
 Consider a GraphQL operation like the following:
 
@@ -81,7 +81,7 @@ This would lead to the following [conceptual] tree-shaking after compilation of 
 
 ### Phase 2
 
-In this phase, we will expand on the previous phase by ahead-of-time compiling the resolution of the operations, their field-resolvers, and invocation thereof into JavaScript code. This essentially does away with any need for AST of the operation during execution.
+In this phase, we will expand on the previous phase by ahead-of-time compiling the resolution of the operations, their field-resolvers, and invocation thereof into JavaScript code. This essentially does away with any need for AST of the operation during execution. This means execution will be faster as no more generic lookups and checks need to be performed.
 
 Consider the GraphQL operation shown in the previous phase, typical generic execution (as [described in the specification](https://spec.graphql.org/June2018/#sec-Execution)) would look something like the following recursive pseudo-code:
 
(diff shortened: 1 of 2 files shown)


freiksenet (approved):

packages/supermassive/README.md
```diff
@@ -2,14 +2,43 @@
...
+## What
 
-There are 3 main parts of supermassive - the executor, query annotator and implicit resolver extractor. Executor is the part that actually runs the queries. It takes resolvers object instead of schema and annotated documents instead of normal documents. Query annotator processes query to include type information inside them. It can be ran as part of query extraction stage in Relay Compiler or eg in `@graphitation/graphql-js-tag`. Implicit resolver extractor writes out resolvers for types that are only implicitly defined in GraphQL SDL, like Unions or Input Objects. It generates typescript file with extracted object that can be merged with the rest of the resolvers.
+Our goals are to provide a solution with the following optimizations for when all GraphQL operations (queries, mutations, subscriptions) needed by the user-experiences are statically known at build-time:
```
orta, line 9:
```suggestion
Supermassive's goals are to provide a solution with the following optimizations for when all GraphQL operations (queries, mutations, subscriptions) needed by the user-experiences are statically known at build-time:
```

packages/supermassive/README.md
```diff
@@ -2,14 +2,43 @@
...
+
+- Bundle size of production targets needs to be as small as possible.
+- Performance is favored over runtime validation.
+
+Such is the case in our scenarios where our schema lives entirely in the client application and some of these applications have very little or no GraphQL needs [yet] other than to suffice the needs of a single or few components.
+
+## Why
+
+Running a GraphQL executor can be an expensive exercise. The JavaScript community has the good fortune of having the official reference implementation of the GraphQL specification being implemented in JavaScript, namely [graphql-js](http://github.com/graphql/graphql-js). Inevitably this means that most general purpose GraphQL libraries in the JavaScript ecosystem end up wrapping it or otherwise rely on it. However, its goal is specifically to be an _all-encompassing_ implementation used for reference needs, _not_ to be an optimized solution for specific use-cases.
```
orta, line 18:
```suggestion
Running a GraphQL executor can be an expensive exercise. The JavaScript community has the good fortune of having the official reference implementation of the GraphQL specification being implemented in JavaScript, namely [graphql-js](http://github.com/graphql/graphql-js). Inevitably this means that most general purpose GraphQL libraries in the JavaScript ecosystem end up wrapping it or otherwise rely on it. However, graphql-js' goal is specifically to be an _all-encompassing_ implementation used for reference needs, _not_ to be an optimized solution for specific use-cases.
```

packages/supermassive/README.md
```diff
@@ -2,14 +2,43 @@
...
+
+## How
+
+Consider a GraphQL schema. It is typically a sizeable chunk of data, both in terms of type/field definition metadata as well as associated field-resolvers and _their_ code dependencies. Statically knowing all GraphQL operations allows us to reduce the bundle size to a minimum by [tree-shaking](https://en.wikipedia.org/wiki/Tree_shaking) all the definition metadata not required by any of the given operations. A JavaScript code bundler can in turn ensure only that code which is needed by the remaining field-resolvers is included in the production bundle. This means that the process is entirely dependency driven by needs expressed by the user-experiences, rather than requiring blunt manual configuration.
+
+Similarly, the GraphQL operations themselves, described using e.g. [GraphQL SDL](https://graphql.org/learn/schema/) or [`graphql-js` AST](https://github.com/graphql/graphql-js/blob/main/src/language/ast.ts), can incur quite some overhead as operations and number of operations grow. Eliminating these from the bundles can save size as well as runtime processing.
```
orta, line 24:
```suggestion
Similarly, running the GraphQL operations themselves, described using e.g. [GraphQL SDL](https://graphql.org/learn/schema/) or [`graphql-js` AST](https://github.com/graphql/graphql-js/blob/main/src/language/ast.ts), can incur quite some overhead as operations and number of operations grow. Eliminating these from the bundles can save size as well as runtime processing.
```

packages/supermassive/README.md
```diff
@@ -1,15 +1,46 @@
...
+
+### Phase 1
+
+In this initial phase, we will achieve the goal of tree-shaking the schema definitions. We do this by inlining required metadata into the documents that describe the operations, after which they can be executed with the need of the entire schema.
+
+### Phase 2
+
+In this phase, we will expand on the previous phase by ahead-of-time compiling the resolution of the operations, their field-resolvers, and invocation thereof into JavaScript code. This essentially does away with any need for AST of the operation during execution.
+
+### Phase 4
```
orta:
```suggestion
### Phase 3
```

alloy, line 36: “There is no step 3 ✨ ” 🤦‍♂️

packages/supermassive/README.md
```diff
@@ -1,15 +1,46 @@
...
+
+Similarly, the GraphQL operations themselves, described using e.g. [GraphQL SDL](https://graphql.org/learn/schema/) or [`graphql-js` AST](https://github.com/graphql/graphql-js/blob/main/src/language/ast.ts), can incur quite some overhead as operations and number of operations grow. Eliminating these from the bundles can save size as well as runtime processing.
+
+## Roadmap
+
+### Phase 1
+
+In this initial phase, we will achieve the goal of tree-shaking the schema definitions. We do this by inlining required metadata into the documents that describe the operations, after which they can be executed with the need of the entire schema.
+
```
orta, line 31: In my head its a little hard to grasp the difference between this and the persisted queries step 3. I think it's effectively dropping bits of the schema used during dev (which you could show as a before/after)

packages/supermassive/README.md
```diff
@@ -1,15 +1,46 @@
...
+### Phase 2
+
+In this phase, we will expand on the previous phase by ahead-of-time compiling the resolution of the operations, their field-resolvers, and invocation thereof into JavaScript code. This essentially does away with any need for AST of the operation during execution.
```
orta, line 34: You could expand with a metaphor like pre-compiling TypeScript to JS ahead of time, so that a server doesn't have to do that work maybe

packages/supermassive/README.md
```diff
@@ -1,15 +1,46 @@
...
+
+### Phase 4
+
+In this final phase, we will make it possible to replace the operations at runtime using a simple identifier, thus allowing GraphQL clients to execute their operations using these identifiers that they obtain through a concept known as ["persisted queries"](https://relay.dev/docs/guides/persisted-queries/). GraphQL clients that do _not_ require graphql-js AST themselves to operate, such as Relay, will benefit from this by being able to reduce their bundle size by stripping out the operations.
```
orta, line 38: IMO, I think you could maybe do a "before and after" here, e.g.

Before: A client sends

```
{ a { b { c, d } } } 
```

After: sends a unique ID which is internally known as that query on the "server'

```
PTAEHUFMBsGMHsC2lQBd5
```

packages/supermassive/README.md
alloy: It isn't just when running them, though. Just the fact of having them in the bundle makes them take up space. Or am I reading/writing it wrong?

packages/supermassive/README.md
alloy: Doh, absolutely. That had been my very first want, and then forgot after writing all this 🤦

alloy: Which do you think works better? I'm leaning towards the diff version:

Screenshot 2021-12-08 at 16 55 26

Screenshot 2021-12-08 at 17 00 05

packages/supermassive/README.md
orta: Yeah, I think this can be rejected now that there's that screenshot in the stage 1 below - if you didn't quite get it here you would by the time you've seen that

packages/supermassive/README.md
alloy: Going with the diff.

packages/supermassive/README.md
alloy: Done

packages/supermassive/README.md
alloy: Yup :D

packages/supermassive/README.md
alloy: I feel like the analogy doesn't really help enough, as TS to JS is explicitly more about transpiring then it is about compiling to something else. Hopefully the examples illustrate the idea enough.

PR #63, opened 2021-12-13
Add graphql-js notice

freiksenet:


alloy:
I want to be a bit more explicit about the license and copyright as to which files it exactly applies to. Can you add a file header to each file like this one? https://github.com/microsoft/graphitation/blob/c30fd1b2b70b8b08d4ea5a6a0d2065f5593311c4/packages/graphql-eslint-rules/src/missing-apollo-key-fields.ts#L1-L4

alloy:
As discussed offline, if you think we will never make changes to any of those files then I can concede with a single notice file, but move it to the jsutils dir and have it mention that it applies to the files in that dir.

Also would be nice to include the version/commit that it was imported from, just for bookkeeping

freiksenet:
@alloy added a notice to the dir. I'll keep notice in root for now, because the work is still a derivative work as a whole.

alloy (approved):
Gotcha, that’s a good point 👍

PR #66, opened 2021-12-16
[ARRDT] Add usePaginationFragment

alloy:


alloy:
I would ideally switch the conversations and messages connections in pagination direction, so to resemble how we actually use them in Teams.

packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks.ts
```diff
@@ -166,10 +170,26 @@ export function useCompiledFragment(
   return data;
 }
 
+interface RefetchOptions {
+  onCompleted?: (error: Error | null) => void;
+  fetchPolicy?: "network-only" | "no-cache";
+  /**
+   * Returns the fetched data. This does not exist in the Relay API.
+   */
+  UNSTABLE_onCompleted?: (
```
alloy, line 179: Should rename this to reflect the difference perhaps

alloy: And comment that this should not be exported

packages/relay-compiler-language-graphitation/src/compilerTransforms/retainConnectionDirectiveTransform.ts
```diff
@@ -0,0 +1,65 @@
+import { IRTransform } from "relay-compiler/lib/core/CompilerContext";
+import { Directive, LinkedField } from "relay-compiler/lib/core/IR";
+import { visit } from "relay-compiler/lib/core/IRVisitor";
+
+type PathWithConnectionDirective = [any[], Directive];
+
+export function retainConnectionDirectiveTransform(
```
alloy, line 7: Add comment about need for this retaining of the directive, which is because relay-compiler by default will strip out all of its client-specific directives. Which normally makes sense, as these should not be sent to the schema, but in our case we want to keep this directive as we will send it to Apollo Client instead.

alloy:
Ready for merging back into the larger feature branch.

PR #76, opened 2021-12-28
[relay-compiler-language-graphitation] optimize AST size

vladar:
# Description

Support [@graphql-tools/optimize](https://github.com/ardatan/graphql-tools/tree/master/packages/optimize) for smaller AST size. By default optimizers will: remove descriptions, remove empty optional properties (like `arguments: []` or `directives: []`) and remove locations if they exist.

One minor issue with this package is that each optimizer does a separate `visit` pass over a document, so it will perform 3 additional passes. I think it's OK-ish for build-time tools though.

### DIFF
--- a/packages/relay-compiler-language-graphitation/package.json
+++ b/packages/relay-compiler-language-graphitation/package.json
@@ -21,12 +21,12 @@
     "just": "monorepo-scripts"
   },
   "dependencies": {
+    "@graphql-tools/optimize": "^1.1.1",
     "relay-compiler-language-typescript": "^14.0.0",
     "typescript": "^4.2.3"
   },
   "devDependencies": {
     "@graphitation/graphql-js-tag": "^0.5.4",
-    "@graphql-tools/optimize": "^1.1.1",
     "@types/dedent": "^0.7.0",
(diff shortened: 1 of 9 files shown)


packages/relay-compiler-language-graphitation/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "devDependencies": {
     "@graphitation/graphql-js-tag": "^0.5.4",
+    "@graphql-tools/optimize": "^1.1.1",
```
alloy, line 29: This will need to be a runtime dependency.

alloy (changes requested):
Great, thanks! Agreed on the multiple visitation passes being fine for build-time tools.

Could you please run `yarn relay` from the example app and add a commit that shows the changes it makes to the artefacts?

packages/relay-compiler-language-graphitation/package.json
vladar: Yeah, it's an oversight from me. Fixed now.

alloy (approved):
Nice, thanks! 🚀

PR #77, opened 2022-01-05
[example] Configure source-maps for React Render Tracker to open editor

alloy:
This sets up React Render Tracker in the example to work for anyone.

@lahmatiy Do you think this is an acceptable generic solution that we can add to your README?


lahmatiy:
LGTM. However, keep in mind that this will work for VS Code users only.

alloy:
> LGTM. However, keep in mind that this will work for VS Code users only.

Fair point 👍  I guess for this example app I am going to make that assumption.

PR #81, opened 2022-01-06
Jvejr/example relay compiler

vejrj:


examples/supermassive-todomvc/package.json
```diff
@@ -3,9 +3,11 @@
   "main": "index.js",
   "version": "1.0.0",
   "scripts": {
+    "start": "concurrently 'yarn generate-graphql-interfaces --watch' 'yarn serve --progress --color --mode=development'",
     "build": "webpack --mode production --config webpack.config.ts",
-    "start": "webpack serve --progress --color --mode=development  --config webpack.config.ts",
-    "extract-schema": "supermassive extract-schema ./src/schema/typeDefs.graphql"
+    "serve": "webpack serve --config webpack.config.ts",
+    "extract-schema": "supermassive extract-schema ./src/schema/typeDefs.graphql",
+    "generate-graphql-interfaces": "relay-compiler --language graphitation --schema ./src/schema/typeDefs.graphql --src ./src"
```
alloy, line 10:
```suggestion
    "relay": "relay-compiler --language graphitation --schema ./src/schema/typeDefs.graphql --src ./src"
```

Let's keep this simple for now. This is also consistent with what I've been calling these tasks in my branch, so when/if we want to rename them we can easily do so in a single sweep.

examples/supermassive-todomvc/package.json
```diff
@@ -3,9 +3,11 @@
   "main": "index.js",
   "version": "1.0.0",
   "scripts": {
+    "start": "concurrently 'yarn generate-graphql-interfaces --watch' 'yarn serve --progress --color --mode=development'",
```
alloy, line 6:
```suggestion
    "start": "concurrently 'yarn relay --watch' 'yarn serve --progress --color --mode=development'",
```

As per below.

examples/supermassive-todomvc/src/components/Todo.tsx
```diff
@@ -2,6 +2,7 @@ import React from "react";
 
 import { useFragment } from "@graphitation/apollo-react-relay-duct-tape";
 import { graphql } from "@graphitation/graphql-js-tag";
+import { TodoFragment$key as TodoFragmentType } from "./__generated__/TodoFragment.graphql";
```
alloy, line 5: I'd rather not rename as it's not really true that this _is_ the type of the fragment data itself, rather it's an opaque fragment reference prop.

examples/supermassive-todomvc/src/components/TodoList.tsx
```diff
@@ -3,6 +3,7 @@ import React from "react";
 import { useLazyLoadQuery } from "@graphitation/apollo-react-relay-duct-tape";
 import { graphql } from "@graphitation/graphql-js-tag";
 import Todo, { TodoFragment } from "./Todo";
+import { TodoListQuery as TodoListQueryType } from "./__generated__/TodoListQuery.graphql";
```
alloy, line 6: This renaming is only necessary right now because of storing the GraphQL doc in a module-scope variable. Instead, can you inline the document at the `useLazyLoadQuery` call-site? Which will be the pattern we want people to actually learn. Then you can remove this renaming too.

examples/supermassive-todomvc/src/components/TodoList.tsx
```diff
@@ -16,7 +17,7 @@ const TodoListQuery = graphql`
 `;
 
 const TodoList = () => {
-  const { data } = useLazyLoadQuery(TodoListQuery, {});
+  const { data } = useLazyLoadQuery<TodoListQueryType>(TodoListQuery, {});
```
alloy, line 20: So replace `TodoListQuery` with ``graphql`...` `` here

alloy (changes requested):

freiksenet (approved):

alloy (approved):
Thanks :shipit:

PR #88, opened 2022-01-06
[ARRDT] Verify that variables with default values and @argumentDefinitions work

### LINKED ISSUE #83: [ARRDT] Add support fragment arguments directives
* https://relay.dev/docs/api-reference/graphql-and-directives/#arguments
* https://relay.dev/docs/api-reference/graphql-and-directives/#argumentdefinitions

### LINKED ISSUE #84: [ARRDT] Record variables with default values in stored request variables

alloy:
Closes #84
Closes #83

The assumption was that we needed to extract default values and persist those in our ‘context’, but that actually didn’t turn out to be true. As long as we don't override a variable in a refetch, the default value will be used as defined in the operation variables definitions. Thus, this PR only demonstrates it working, no implementation changes were made.

### DIFF
--- a/examples/apollo-watch-fragments/src/App.tsx
+++ b/examples/apollo-watch-fragments/src/App.tsx
@@ -14,11 +14,7 @@ const App: React.FC = () => {
 
   const result = useLazyLoadQuery<AppQueryType>(
     graphql`
-      query AppQuery(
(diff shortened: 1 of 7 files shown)


vladar (approved):
Nice! 👍

freiksenet (approved):

PR #92, opened 2022-01-13
Add initial version of website

alloy:
Nothing fancy. Uses Docusaurus and each package can have their section.

Needs some further refining, especially of broken homepage, but for now I need links to hooks docs.


alloy:
Part of #87

PR #100, opened 2022-01-17
[SUPERMASSIVE] added subscribe operation

vejrj:


packages/supermassive/src/__tests__/subscribe.test.ts
```diff
@@ -0,0 +1,281 @@
...
+    },
+  },
+];
+
+describe("subscribeWithSchema", () => {
+  testCases.forEach(({ name, document, variables }: TestCase) => {
+    it(name, async () => {
+      await compareResultsForSubscribeWithSchema(document, variables);
+    });
+  });
```
vladar, line 57: Can we switch to [`test.each` or `it.each` syntax](https://jestjs.io/docs/api#testeachtablename-fn-timeout) - IDEs support it a bit better than custom loops.

change/@graphitation-supermassive-51bd644d-9254-42bb-a0c4-c6653f90b7d1.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "minor",
+  "comment": {
```
freiksenet, line 3: This should be a string comment describing change.

examples/supermassive-todomvc/src/schema/link.ts
```diff
@@ -1,10 +1,17 @@
-import { execute as graphQLExecute, isInputType, parse } from "graphql";
+import {
```
freiksenet, line 1: [nit] unused imports

packages/supermassive/src/subscribeWithoutSchema.ts
```diff
@@ -0,0 +1,286 @@
...
+
+import {
+  ObjectTypeResolver,
+  FunctionFieldResolver,
+  FieldResolver,
+  Resolvers,
+  ExecutionResult,
+} from "./types";
+
+export interface SubscriptionArgs {
```
freiksenet, line 36: Are those different from execute args? Seems they are the same in graphql-js, let's unify it here too.

packages/supermassive/src/subscribeWithoutSchema.ts
```diff
@@ -0,0 +1,286 @@
...
+      name: {
+        kind: Kind.NAME,
+        value: "String",
+      },
+    };
+  } else {
+    returnTypeNode = fieldNodes[0].__type as TypeNode;
+    returnTypeName = typeNameFromAST(returnTypeNode);
+    const typeResolvers = exeContext.resolvers[typeName];
+    resolveFn = ((typeResolvers as
```
freiksenet, line 238: Are those any casts still needed? I think the type should support this now.

freiksenet (commented):
Looking good, some really minor issues. Great work!

packages/supermassive/src/subscribeWithoutSchema.ts
vejrj: We added `subscribeFieldResolver`

change/@graphitation-supermassive-51bd644d-9254-42bb-a0c4-c6653f90b7d1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "[SUPERMASSIVE] added subscribe operation",
```
freiksenet, line 3:
```suggestion
  "comment": "Added support for subscribe operation",
```

packages/supermassive/src/subscribeWithoutSchema.ts
freiksenet: You can add it to all ExecutionArgs, like in graphql-js.

https://github.com/graphql/graphql-js/blob/main/src/execution/execute.ts#L154

freiksenet (approved):

PR #108, opened 2022-01-24
Fix Webpack 4 compat

freiksenet:
Change main format to support webpack 4.

Make default imports for third-party CommonJS dependencies or remove them

### DIFF
--- a/packages/supermassive/src/executeWithSchema.ts
+++ b/packages/supermassive/src/executeWithSchema.ts
@@ -1,5 +1,4 @@
-import { makeExecutableSchema } from "@graphql-tools/schema";
-import { isInputType, parse } from "graphql";
+import { isInputType, buildASTSchema } from "graphql";
 import {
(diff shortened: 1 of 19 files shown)


packages/supermassive/src/executeWithSchema.ts
```diff
@@ -20,7 +19,7 @@ export function executeWithSchema({
   fieldResolver,
   typeResolver,
 }: ExecutionWithSchemaArgs): PromiseOrValue<ExecutionResult> {
-  const schema = makeExecutableSchema({ typeDefs, resolvers });
+  const schema = buildASTSchema(typeDefs);
```
alloy, line 22: I'm unsure why this is included here, was this intentional?

freiksenet: Another react like example that was breaking for office start.

alloy (approved):

PR #111, opened 2022-01-26
Supermassive get argument values fix

vejrj:
When the `variableValues` doesn't contain variableName, which is specified in function arguments AND it isn't `isNonNullType` (it's not required) we just want to skip it


freiksenet (approved):
Great find.

PR #115, opened 2022-02-02
[ARRDT] Publish fixes

alloy:


alloy:
Haven't yet been able to debug `publishConfig` and why it doesn't seem to get applied when I run `npm publish`.

vladar (approved):

scripts/just.config.ts
```diff
@@ -40,7 +40,19 @@ export const build = () => {
...
+                  const absolutePath = path.resolve(
+                    args.importer,
+                    "..",
+                    extPath
+                  );
+                  if (
+                    fs.existsSync(absolutePath) &&
+                    fs.lstatSync(absolutePath).isDirectory()
+                  ) {
+                    extPath = extPath + path.sep + "index.mjs";
```
vladar, line 52: After some thinking, we may prefer to use `/` here vs `path.sep`. Otherwise there could be potential problems with character escaping depending on how this will be used later. Windows should handle the `/` just fine.

alloy: Yup, right so you are 👍

Issue #126, opened 2022-02-15
[ARRDT] Make node field policy work for ids without typename prefix


alloy:
When node ids are truly unique, Apollo Cache can be configured to not prefix typename. In this case, we need to adjust the cache redirect from the node field policy, which means we can do less work as we don't need to infer the typename from the fragment.

Issue #113, opened 2022-01-28
[ARRDT] look into optimism thrashing


alloy:
According to @vladar any read from the cache adds entries to optimism. We need to verify this and try to avoid it, perhaps as described here https://github.com/apollographql/apollo-client/issues/9306#issuecomment-1015634246

vladar:
Yes, I see optimism cache growing on simple clicks on todo items:

![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/115628/152379815-8aad7eb1-2c04-403a-861c-9648c27047bd.png)

vladar:
One offender is this call where we create a new fragment document object every time (note: `fragment: {` is always a new object): 

https://github.com/microsoft/graphitation/blob/cc843406967da7c964bd5c68e79e1cc3f1b234f5/packages/apollo-react-relay-duct-tape/src/storeObservation/nodeFromCacheFieldPolicy.ts#L42-L52

A simple caching of this document in a `WeakMap` (using `options.query` as a key) allowed to lower growth rate to 2 new entries per click vs 4.

Those 2 remaining items are still to be investigate - I don't think they should appear at all.

vladar:
One thing that popped up during a discussion: graphitation compiler may produce multiple instances of fragment documents (one in queries, and another for fragment hooks) which may end up in even larger number of optimism entries.

We need to investigate if it actually affects optimism size.

alloy:
> One offender is this call where we create a new fragment document object every time (note: `fragment: {` is always a new object):
> 
> https://github.com/microsoft/graphitation/blob/cc843406967da7c964bd5c68e79e1cc3f1b234f5/packages/apollo-react-relay-duct-tape/src/storeObservation/nodeFromCacheFieldPolicy.ts#L42-L52

This one is resolved by #127.

PR #127, opened 2022-02-15
[ARRDT] Use Apollo cache redirect instead of reading from cache ourselves

alloy:
Related to #113 

Follow-up work defined in https://github.com/microsoft/graphitation/issues/126


alloy:
Going for the merge

PR #131, opened 2022-03-01
Force parser not to generate locations for AST nodes

sergey-stoyan:
This change is required because of Apollo 3.5.x, which isn't able to dedupe queries that contain location data.


alloy:
Interesting. This seems fine, less data the better for our bundles, but as discussed in-person, we need to be aware that the assumptions made about AST data by Apollo lib isn't consistent [with graphql-js AST typing, at least]. To iterate:

- Apollo Client doesn't dedupe queries that contain location data
- graphql-tag still expects `loc.source` to exist for interpolated fragments https://github.com/apollographql/graphql-tag/blob/f2de37aa4396834e00bd08db1fa710ad7b365ada/src/index.ts#L127

alloy (approved):

PR #136, opened 2022-03-08
[supermassive] Add deps to types task

alloy:


dannyvv (approved):

alloy:
@vejrj @freiksenet Can/should we check these artefacts in, instead of requiring them to be build before we can invoke some of the other tasks?

PR #137, opened 2022-03-08
Exclude generated files from linting

dannyvv:
On machines with many cores, `lint` and `test` can run in parralel for `@graphitation/supermassive` project.

This makes sures that __generated__ folders are excluded from lint. As they are part of `.gitignore` already.


alloy (approved):
Thanks 🙏

PR #138, opened 2022-03-08
adding azure pipeline to do the release instead

kenotron:
1. added a release pipeline for ado
2. removed the pipeline from github actions

compliance-job template is derived from: https://github.com/microsoft/fluentui/blob/f2ecf6614a11aaf38d2d66bb0f61143487b74e6d/.devops/templates/compliance-job.yml

and the release.yml is heavily influenced by what was already there.


kerrynf (approved):

.azure-devops/templates/compliance-job.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - task: AssetRetention@3
+        displayName: 🧭 Arrow Retention
+        inputs:
+          ArrowServiceConnection: "Arrow_domoreexp_Teamspace_PROD"
+          AssetGroupName: "$(System.TeamProject)_$(Build.DefinitionName)"
+          AssetNumber: "$(Build.BuildId)"
+          IsShipped: false
+          DropsToRetain: "CodeAnalysisLogs"
+        condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
```
alloy, line 48: ?

```suggestion
        condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/main'))
```

.azure-devops/templates/compliance-job.yml
```diff
@@ -0,0 +1,58 @@
...
+        displayName: "Use .NET Core sdk 3.x"
+        inputs:
+          version: 3.x
+          steps:
+
+      - task: skvso.servicetree-build-tasks.servicetree-link-build-task.servicetree-link-build-task@1
+        displayName: "ServiceTree Integration"
+        inputs:
+          ServiceTreeGateway: "ServiceTree Gateway"
+          Service: 6F8CD842-E117-412F-BAE4-56A3B6166594
```
alloy, line 21: Assuming this is graphitation's ID in the service tree? If this config is meant to serve as a template it would be good to call that out or change it to a variable defined at the start of the config?

alloy (approved):
Thanks for doing this @kenotron 🙏 

One question about the branch ref, but otherwise LGTM 👍

.azure-devops/templates/compliance-job.yml
kenotron: good call!

.azure-devops/templates/compliance-job.yml
kenotron: I think the right thing to do is to add a parameter to the template... lemme add that - this is a copy-pasta from fluentui :)

.azure-devops/templates/compliance-job.yml
kenotron: fixed!

.azure-devops/templates/compliance-job.yml
kenotron: OKAY this is fixed!

PR #139, opened 2022-03-16
Creates a compliant release per executive order

kenotron:
In this PR:
1. we added the release steps inside the azure pipelines.yml
2. updated which pool to use that is "compliant"
3. made sure the secrets are also coming from a compliant source

Notes:
I refactored the jobs a bit since both jobs need the servicetree task to be run, so I've decided to factor that task out.

What I cannot test in this: the git push part of beachball & the npm publish. These would presumably work since the tokens are already setup...


alloy (approved):
:shipit:

kenotron:
Okay this is good to go now - validated working in the ADO pipeline side!

PR #140, opened 2022-03-24
[GRAPHQL_ESLINT_RULE] Added an eslint rule for enforcing naming conve…

vejrj:
…ntions

### DIFF
--- a/packages/graphql-eslint-rules/src/__tests__/fragment-naming-conventions.test.ts
+++ b/packages/graphql-eslint-rules/src/__tests__/fragment-naming-conventions.test.ts
@@ -1,8 +1,3 @@
-/*
- * Taken from https://github.com/dotansimha/graphql-eslint/blob/300f73be802bdd06432a5df34939521d1ce0d93b/packages/plugin/tests/require-id-when-available.spec.ts
- * MIT license https://github.com/dotansimha/graphql-eslint/blob/300f73be802bdd06432a5df34939521d1ce0d93b/LICENSE
- */
-
 import { TextDecoder } from "util";
 global.TextDecoder = TextDecoder as any;
 
--- a/packages/graphql-eslint-rules/src/__tests__/fragment-naming-conventions.test.ts
+++ b/packages/graphql-eslint-rules/src/__tests__/fragment-naming-conventions.test.ts
@@ -1,0 +1,97 @@
+/*
+ * Taken from https://github.com/dotansimha/graphql-eslint/blob/300f73be802bdd06432a5df34939521d1ce0d93b/packages/plugin/tests/require-id-when-available.spec.ts
+ * MIT license https://github.com/dotansimha/graphql-eslint/blob/300f73be802bdd06432a5df34939521d1ce0d93b/LICENSE
+ */
+
+import { TextDecoder } from "util";
+global.TextDecoder = TextDecoder as any;
+
+import {
+  GraphQLRuleTester,
+  ParserOptions,
+} from "@graphql-eslint/eslint-plugin";
+import rule from "../fragment-naming-conventions";
+
+const ruleTester = new GraphQLRuleTester();
+
+const TEST_SCHEMA = /* GraphQL */ `
+  type User {
+    id: ID!
+    name: String
+  }
+`;
+
+const WITH_SCHEMA = {
+  parserOptions: <ParserOptions>{
+    schema: TEST_SCHEMA,
+  },
+};
+
+const errorMessage =
+  "Fragment should follow the naming conventions, the expected name is GraphqlEslintRulesUserFragment OR GraphqlEslintRulesUserFragment_optionalSuffix_anotherOptionalSuffix";
+
+ruleTester.runGraphQLTests(
+  "fragment-naming-conventions",
+  rule as any, // FIXME: Not casting this as any leads to ts(2590) error
+  {
+    valid: [
+      {
+        ...WITH_SCHEMA,
+        filename: "src/user-query.graphql",
+        code: `fragment GraphqlEslintRulesUserFragment on User { id name }`,
+      },
+      {
+        ...WITH_SCHEMA,
+        filename: "src/user-query.graphql",
+        code: `fragment GraphqlEslintRulesUserFragment_optionalSuffix on User { id name }`,
+      },
+      {
+        ...WITH_SCHEMA,
+        filename: "src/user-query.graphql",
+        code: `fragment GraphqlEslintRulesUserFragment_optionalSuffix_anotherOptionalSuffix on User { id name }`,
+      },
+    ],
+    invalid: [
+      {
+        ...WITH_SCHEMA,
+        filename: "src/user-query.graphql",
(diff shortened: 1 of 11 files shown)


packages/graphql-eslint-rules/src/__tests__/fragment-naming-conventions.test.ts
```diff
@@ -0,0 +1,97 @@
+/*
+ * Taken from https://github.com/dotansimha/graphql-eslint/blob/300f73be802bdd06432a5df34939521d1ce0d93b/packages/plugin/tests/require-id-when-available.spec.ts
+ * MIT license https://github.com/dotansimha/graphql-eslint/blob/300f73be802bdd06432a5df34939521d1ce0d93b/LICENSE
+ */
```
alloy, line 4: I don't think this is still needed for what you have below. The code below is basically an entire new test case that happens to use the same rule tester API.

packages/graphql-eslint-rules/src/fragment-naming-conventions.ts
```diff
@@ -0,0 +1,129 @@
...
+    node: newNode,
+    message: `Fragment should follow the naming conventions, the expected name is ${expectedName} OR ${expectedName}_optionalSuffix_anotherOptionalSuffix`,
+  });
+}
+
+const rule: GraphQLESLintRule = {
+  meta: {
+    type: "problem",
+    docs: {
+      category: "Operations",
```
alloy, line 62: Is this an ESLint thing or is this supposed to say “Fragments”?

packages/graphql-eslint-rules/src/fragment-naming-conventions.ts
```diff
@@ -0,0 +1,129 @@
...
+      end: {
+        line: node.loc.end.line,
+        column: node.loc.end.column - 1,
+      },
+    },
+  };
+
+  context.report({
+    node: newNode,
+    message: `Fragment should follow the naming conventions, the expected name is ${expectedName} OR ${expectedName}_optionalSuffix_anotherOptionalSuffix`,
```
alloy, line 54: Why the `_anotherOptionalSuffix` part? Is 2 underscores we specifically want to suggest in some scenarios? If not, then why stop at 2? I.e. it becomes arbitrary.

packages/graphql-eslint-rules/src/fragment-naming-conventions.ts
```diff
@@ -0,0 +1,129 @@
...
+    docs: {
+      category: "Operations",
+      description: `Enforce more descriptive fragment names`,
+      requiresSiblings: true,
+      examples: [
+        {
+          title: "Incorrect",
+          code: /* GraphQL */ `
+            # packages/eslint-rules-example/foo.query.graphql
+            fragment ExampleFoo on User {
```
alloy, line 70: It's better to not use too generic names in examples, but rather something the reader can relate to. E.g.

```suggestion
            fragment Image on User {
```

(Also requires updating of path.)

packages/graphql-eslint-rules/src/fragment-naming-conventions.ts
```diff
@@ -0,0 +1,129 @@
...
+              id
```
alloy, line 71: …and then replace this too:

```suggestion
              url
```

packages/graphql-eslint-rules/src/fragment-naming-conventions.ts
```diff
@@ -0,0 +1,129 @@
...
+            }
+          `,
+        },
+        {
+          title: "Correct",
+          code: /* GraphQL */ `
+            # packages/eslint-rules-example/foo.query.graphql
+            fragment EslintRulesExampleFooFragment_optionalSuffix_anotherSuffix on User {
```
alloy, line 79: …and then:

```suggestion
            fragment MsTeamsAvatarImage on User {
```

I would suggest leaving the suffix off entirely here. Until we introduce the Relay pattern, people won't know what to do with it and it may confuse them.

packages/graphql-eslint-rules/src/operation-naming-conventions.ts
```diff
@@ -0,0 +1,126 @@
...
+          title: "Correct",
+          code: /* GraphQL */ `
+            # packages/eslint-rules-example/foo.query.graphql
+            query EslintRulesExampleFooQuery {
+              user {
+                id
+              }
+            }
+          `,
+        },
```
alloy, line 76: Same feedback on examples applies here too.

packages/graphql-eslint-rules/src/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  const pathname = join(dir, "package.json");
+  let stat;
+
+  try {
+    stat = fs.statSync(pathname, { throwIfNoEntry: false } as StatOptions);
+  } finally {
+    if (stat && stat.isFile()) {
+      return pathname;
+    }
+  }
```
alloy, line 14: For my learning, what could it still throw for?

packages/graphql-eslint-rules/src/utils.ts
```diff
@@ -0,0 +1,22 @@
+import { join, dirname } from "path";
+import fs, { StatOptions } from "fs";
+
+export function checkDirForPkg(dir: string): string | null {
+  const pathname = join(dir, "package.json");
+  let stat;
+
+  try {
+    stat = fs.statSync(pathname, { throwIfNoEntry: false } as StatOptions);
```
alloy, line 9: Can you move the type assertion to the variable declaration instead?

alloy (commented):
Nice 👌

Just some small comments.

alloy:
Btw, should there be an auto-fix?

packages/graphql-eslint-rules/src/utils.ts
vejrj: I tried to run it in TMP and for some reason it threw an error even with `throwIfNotEntry`

packages/graphql-eslint-rules/src/fragment-naming-conventions.ts
vejrj: It has pre-defined type in graphql-eslint. It can be either Operation or Schema https://github.com/B2o5T/graphql-eslint/blob/4bcee4a7849602076fb9e0d9ab33e78141e20a8e/packages/plugin/src/types.ts#L51

alloy: Gotcha.

alloy: I guess they mean execution doc vs schema definition doc (?)

freiksenet:
So this does no checks for .js/.ts files yet, does it? It is needed eg for Nova component and future relay patterns.

vejrj:
> So this does no checks for .js/.ts files yet, does it? It is needed eg for Nova component and future relay patterns.

No, it doesn't. It will be a different rule.

alloy (approved):

PR #150, opened 2022-03-31
[ado] Pierce ADO NPM feed cache for published packages

alloy:
Still need to add the PAT to the secrets group, looking into ADO CLI now.

### DIFF
--- a/.azure-devops/graphitation-release.yml
+++ b/.azure-devops/graphitation-release.yml
@@ -8,6 +8,8 @@ variables:
     value: production,externalfacing
   - name: serviceTreeID
     value: 6F8CD842-E117-412F-BAE4-56A3B6166594
+  - name: adoNpmFeedBaseUrl
+    value: https://pkgs.dev.azure.com/domoreexp/_apis/packaging/feeds/npm-mirror
 
 jobs:
   - job: compliance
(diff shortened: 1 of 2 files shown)


.azure-devops/graphitation-release.yml
```diff
@@ -41,3 +43,7 @@ jobs:
         displayName: Configure git for release
       - script: yarn release -y -n $(ossNpmToken) --access public
         displayName: Release
+      - template: ./steps/pierce-ado-npm-mirror-cache.yml
+        parameters:
+          adoNpmFeedPat: $(adoNpmFeedPat)
```
alloy, line 48: Is this the correct way to make a secret available to a template?

vejrj (approved):

alloy:
@kenotron I'd also like to make sure this step does not in any way fail the pipeline, as it needs some testing and isn't a required step. Do you know how I can ensure that?

.azure-devops/graphitation-release.yml
kenotron: try just using "ossNpmToken" here.

alloy: @kenotron So from our offline chat I assume you now understand that this isn't what we're after.

kenotron: I'll add this for you :)

alloy:
At least in a test with no publishing it doesn’t fail and if it were to fail after the next publish, it would at least fail _after_ the publish.

PR #151, opened 2022-03-31
[ARRDT] Support for strict Global Object Identification setup

alloy:
When an application has been configured to not prefix store keys with the typename, because the id values follow the Global Object Identification spec, this needs to be taken into account when interacting with the store.


packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledPaginationFragment.ts
```diff
@@ -88,9 +95,10 @@ function useLoadMore({
               "usePaginationFragment(): Expected mainFragment metadata"
             );
             const cacheSelector: DataProxy.Fragment<any, any> = {
-              id: fragmentReference.id
-                ? `${mainFragment.typeCondition}:${fragmentReference.id}`
-                : "ROOT_QUERY",
+              id: cache.identify({
+                __typename: mainFragment.typeCondition,
+                id: fragmentReference.id as StoreValue,
+              }),
```
alloy, line 101: This here is the actual change.

alloy:
Also added some CI/publishing related changes. Hopefully we can get back to releasing these `next` versions in an automated fashion as well.

alloy:
Pushed some changes to fix TS dep issues we have. Will PR that change to `main` separately in a follow-up.

vladar (approved):
Approving the main change (didn't actually review the CI setup - just skimmed through)

PR #152, opened 2022-04-01
HOTFIX: fragment-naming-conventions file checks the filename conventions AND uses this filename for checking the fragment name instead of building in from the lastDirectory name

vejrj:
Fixing a bug in the `fragment-naming-conventions` rule. The rule was checking/autofixing the name in the following way.
[LastDirectoryPrefix][PascalCasedFileName]Fragment. Nonetheless, [PascalCasedFileName] was fixed by the `operation-naming-conventions` rule to contain lastDirectory prefix too, so as result the lastDirectoryPrefix was autofix twice =>
[LastDirectoryPrefix][LastDirectoryPrefix][PascalCasedFileNameWithoutPrefix]Fragment


alloy (approved):
:shipit:

alloy:
Will be very interesting to see if the ADO cache piercing will work as intended 🤞

PR #154, opened 2022-04-01
supermassive packages fixes

alloy:
In [current](https://github.com/microsoft/graphitation/blob/dfa894cf781629179c5905adb8fb263035a07e3f/packages/supermassive/package.json#L53) `main` the TS dep is listed as ^4.4.3 , but then it uses a compiler API, which was changed in a breaking manner in [TS 4.5](https://github.com/microsoft/TypeScript/pull/45998), but it uses it in the old manner—i.e. [it passes 2 arguments](https://github.com/microsoft/graphitation/blob/dfa894cf781629179c5905adb8fb263035a07e3f/packages/supermassive/src/extractImplicitTypesToTypescript.ts#L132-L135) instead of 3. For now pinning supermassive to < 4.5, but we should probably be consistent and on latest.

### DIFF
--- a/examples/supermassive-todomvc/package.json
+++ b/examples/supermassive-todomvc/package.json
@@ -1,6 +1,7 @@
 {
   "name": "supermassive-todomvc",
   "main": "index.js",
+  "private": true,
   "version": "1.0.0",
   "scripts": {
     "start": "concurrently 'yarn relay --watch' 'yarn serve --progress --color --mode=development'",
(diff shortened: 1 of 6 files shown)


examples/supermassive-todomvc/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "supermassive-todomvc",
   "main": "index.js",
+  "private": true,
```
alloy, line 4: CI just pushed this package to public npm 🙈

PR #173, opened 2022-05-13
[payload-generator] Rewrite to use graphql-js executor and make feature complete

alloy:
This now passes all tests for features we support in duct-tape.

We do deviate in a few places, most notably we do not require the `@relay_test_operation` directive to generate complete payloads. As this lib should be useful for anyone using graphql-js, I don't quite feel like degrading this port specifically for relay compatibility. (Instead I think we'll end up inserting the directive in every one of our tests with a codemod, when needed.)

There's some cleanup/refactor work left to be done, which I might do before cutting v1, but for now I want a published version we can use in TMP.


patches/relay-test-utils-internal+11.0.2.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/node_modules/relay-test-utils-internal/lib/testschema.graphql b/node_modules/relay-test-utils-internal/lib/testschema.graphql
+index ce0e0f3..1071e39 100644
+--- a/node_modules/relay-test-utils-internal/lib/testschema.graphql
++++ b/node_modules/relay-test-utils-internal/lib/testschema.graphql
```
vladar, line 4: I assume this file is not needed for bundled package? Just want to make sure that the bundled package won't break because of this patch.

vladar (approved):

patches/relay-test-utils-internal+11.0.2.patch
alloy: Correct, this is only needed for our tests.

PR #181, opened 2022-06-02
Apollo devtools migration

vejrj:


vejrj:
> I think the package should be renamed to `apollo-devtools` as it won't work for general graphql case, only for apollo client-specific implementations (unless I am missing something).

renamed

vladar (approved):

PR #184, opened 2022-06-08
Graphql codegen supermassive typed document node plugin

vejrj:


packages/graphql-codegen-supermassive-typed-document-node-plugin/README.md
```diff
@@ -0,0 +1,5 @@
+# Custom GraphQL codegen plugins
+
+## supermassive-typed-document-node
+
+Do the same stuff as the original typed-document-node plugin except it extends GraphQL AST data to match Supermassive AST
```
alloy, line 5: Always try to reduce work on the reader and include links to references. i.e. the original plugin and perhaps supermassive AST details.

packages/graphql-codegen-supermassive-typed-document-node-plugin/src/index.ts
```diff
@@ -0,0 +1,84 @@
+/*
+ * Taken from https://github.com/dotansimha/graphql-code-generator/blob/4fee8c8c523b30163e913438b85a064c58e39087/packages/plugins/typescript/typed-document-node/src/index.ts
+ * MIT license https://github.com/dotansimha/graphql-code-generator/blob/4fee8c8c523b30163e913438b85a064c58e39087/LICENSE
+ */
```
alloy, line 4: ✨  👌

scripts/just.config.ts
```diff
@@ -13,15 +13,15 @@ import * as glob from "fast-glob";
...
+    outDir: argv().outDir || "lib",
     emitDeclarationOnly: true,
   });
 };
 
 export const build = () => {
   const baseEsbuildOptions: EsbuildBuildOptions = {
     entryPoints: glob.sync(["src/**/*.{ts,tsx}", "!src/**/__tests__/**"]),
-    outdir: "lib",
+    outdir: argv().outDir || "lib",
```
alloy, line 24: Let's not do it this way, consistency in build systems is important. Instead, the pattern typically is that you add a commonjs file that acts as a trampoline.

i.e. add a `index.js` file in the root of the package which does something like the following:

```js
module.exports = require("./lib/index");
```

…and a `index.mjs` file which does:

```js
export * from "./lib/index.mjs"
```

alloy (changes requested):

scripts/just.config.ts
alloy: Actually, isn't this what the `main` and `module` keys are for in the `package.json` file?

packages/graphql-codegen-supermassive-typed-document-node-plugin/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "lint": "monorepo-scripts lint",
+    "types": "monorepo-scripts types",
+    "just": "monorepo-scripts"
+  },
+  "devDependencies": {
+    "@graphitation/supermassive": "*",
+    "@graphql-codegen/plugin-helpers": "^1.18.2",
+    "@graphql-codegen/visitor-plugin-common": "^1.17.20",
+    "@types/jest": "^26.0.22",
+    "graphql-tag": "^2.11.0",
```
alloy, line 22: Do we really need to use this package? We actually have one ourselves that does what we would normally need.

alloy: Let’s follow up on this after your vacation. Can you file a work item?

alloy (approved):

PR #209, opened 2022-08-11
Apollo devtools: [HOTFIX] graphiql storage fixed

vejrj:


packages/rempl-apollo-devtools/src/subscriber/graphiql/graphiql.tsx
```diff
@@ -38,10 +38,31 @@ export const GraphiQLRenderer = React.memo(() => {
...
+
+  return {
+    get length() {
+      return Object.keys(storage).length;
+    },
+    setItem: function (key: string, value: string | null) {
+      storage[key] = value || "";
+    },
+    getItem: function (key: string) {
+      return key in storage ? storage[key] : null;
```
vladar, line 61: Minor nit: there are system fields that exist on prototype and may give false positives (e.g. `"constructor" in {}` is `true`). So maybe use `hasOwnProperty` instead?

```suggestion
      return storage.hasOwnProperty(key) ? storage[key] : null;
```

vladar (approved):

PR #230, opened 2022-10-24
SuperMassive: AST Error Message fix

kerrynf:
When running against an updated package, the compiler generated a message to indicate a field was queried that was unknown: "Unhandled: undefined".

This message is not useful - updating the message to output the name of the field makes it possible to find the query with the field and correctly update schema for the new fields.

It would be even better if there was some more metadata that could be added here - the parent type for the field, or the file in which the field occurred.

### DIFF
--- a/packages/supermassive/src/ast/addTypesToRequestDocument.ts
+++ b/packages/supermassive/src/ast/addTypesToRequestDocument.ts
@@ -52,7 +52,8 @@ export function addTypesToRequestDocument(
           };
           return newNode;
         }
-        throw new Error(`Unhandled: ${type}`);
+        // This happens whenever a new field is requested that hasn't been defined in schema
+        throw new Error(`Unhandled node: ${node.name.value}, type: ${type}`);
       },
     }),
   );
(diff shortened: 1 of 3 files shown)


packages/supermassive/src/ast/addTypesToRequestDocument.ts
```diff
@@ -52,7 +52,8 @@ export function addTypesToRequestDocument(
           };
           return newNode;
         }
-        throw new Error(`Unhandled: ${type}`);
+        // This happens whenever a new field is requested that hasn't been defined in schema
```
freiksenet, line 55: Visit gets the following params.

```export type ASTVisitFn<TVisitedNode extends ASTNode> = (
  /** The current node being visiting. */
  node: TVisitedNode,
  /** The index or key to this node from the parent node or Array. */
  key: string | number | undefined,
  /** The parent immediately above this node, which may be an Array. */
  parent: ASTNode | ReadonlyArray<ASTNode> | undefined,
  /** The key path to get to this node from the root node. */
  path: ReadonlyArray<string | number>,
  /**
   * All nodes and Arrays visited before reaching parent of this node.
   * These correspond to array indices in `path`.
   * Note: ancestors includes arrays which contain the parent of visited node.
   */
  ancestors: ReadonlyArray<ASTNode | ReadonlyArray<ASTNode>>,
) => any;```

Let's add path: ${path.join('.') to the error.

freiksenet (changes requested):

freiksenet:
Thank you!

alloy (approved):

PR #231, opened 2022-10-26
contextPath relative to CWD and resolvers return readonly array

vejrj:


vladar:
I think we should consider having `Readonly` in `PromiseOrValue` type instead. This will make generated types a bit more readable.

[Example in TS playground][1]

@freiksenet @vejrj what do you think?

[1]: https://www.typescriptlang.org/play?ts=4.3.5#code/PTAEDECcHsFtQBYBckAcDOAuEBzAlkggK4BGAdAMZzACyeFM60AZksACoCeqApgMoM8qNnnToiPdMACMAZgCcAJlkBiUeJ5VYsHgDskAWgCsRgOxGADPIXyAUEm49QASW1EkAQxIAbHgAVIPFgCPAA3JwBeUCJdABMeZjxdHljQAB9QXSJvb3TQEmhoXw9dPPQkQN0cPKzYEh5IPPAYiiQ8aF0AbltbHgAPVGhIJFAHXhc3Tx8eAB52AD5QCNtQUHZQfqQ9WPQJ2HcvXwCgkPDQAH410EwVq83t3YBBSEgPThmk5gbQAFVFy9c+ymvmer3ef2ut3W9ziuxoHlQH10X0aAGkADSgT7fABq-z2B2m8MRGNAeMhq2hfS2sNAfB4SCRKNANHxgMJvnpjNZ1wJwJ4AHkSAArTSMhbdXoDIYjMZOdn80FvOaLKIAJR4HliHW8nCV7wVh1mC3m3X6g2Go0cfKNxJmpPJ6s12t0urthum9vmmI9vhmeNNUotsutvv4DJVS1AGq1Os4XJmYZVgfNMqt4zDQtFrUjUQA3qBIM646AANqorGlADWPE4LDWAF1MDbPexyw3FgBfSU9ECgAVERpyrD2a0BOCiQWQHEebwSXMtv0LPLj4LoWZJk09JJbSDMDwUJw0aDxXJ525F2OuzigRI8byxZsxl26-UzZ9xt-lSo4ebzWydj0cq3ng96pFEAAUsAnvezbHqeACUSyLKuk4CtOs7zt+SQ4KWDZ4f+thULo5QgWB4C6M2d4PlGUEwd4SERIs0GnmQ1GxLYQA

freiksenet:
Yeah, this looks good. Let's do that.

packages/supermassive/src/bin/supermassive.ts
```diff
@@ -66,15 +66,11 @@ function getFiles(inputs: Array<string>) {
...
-  outputDir: string,
-  inputPath: string,
-  contextImport: string | undefined,
-) {
+function getContextPath(outputDir: string, contextImport: string | undefined) {
   if (!contextImport) {
     return;
   }
-  const contextDir = path.join(path.dirname(inputPath), contextImport);
+  const contextDir = path.join(process.cwd(), contextImport);
```
alloy, line 73: Just FYI for future changes, this is equivalent to `path.resolve(contextImport)`

packages/supermassive/src/codegen/resolvers.ts
```diff
@@ -201,7 +201,10 @@ function createResolversReducer(
     ListType: {
       leave({ type }): ts.TypeNode {
         return createNullableType(
-          factory.createArrayTypeNode(type as ts.TypeNode),
+          factory.createTypeReferenceNode(
+            factory.createIdentifier("ReadonlyArray"),
+            [type as ts.TypeNode],
+          ),
```
alloy, line 207: iirc, the way to do this with modern TS and through the API is something like:

```ts
factory.createTypeOperatorNode(ts.SyntaxKind.ReadonlyKeyword, factory.createArrayTypeNode(type))
```

Which would produce:

```ts
readonly X[]
```

alloy (approved):

vladar (approved):

packages/supermassive/src/jsutils/PromiseOrValue.ts
```diff
@@ -1 +1,25 @@
...
+  : T extends Set<infer M>
+  ? ImmutableSet<M>
+  : ImmutableObject<T>;
+
+export type ImmutableArray<T> = ReadonlyArray<Immutable<T>>;
+export type ImmutableMap<K, V> = ReadonlyMap<Immutable<K>, Immutable<V>>;
+export type ImmutableSet<T> = ReadonlySet<Immutable<T>>;
+export type ImmutableObject<T> = { readonly [K in keyof T]: Immutable<T[K]> };
+
+export type PromiseOrValue<T> = Immutable<T> | Promise<Immutable<T>>;
```
alloy: I have to wonder how expensive this might be vs the previous inlined solution, when compiling or trying to get instant feedback from the LSP.

vladar:
Generic immutable seems to be a hard thing to do right (see [full discussion in this thread](https://github.com/Microsoft/TypeScript/issues/13923)).

 So, for now we'll have to stick with explicit `ReadonlyArray`. We can revisit it a bit later for our specific case, that doesn't really require a generic solution.

packages/supermassive/src/jsutils/PromiseOrValue.ts
vladar: Yeah, there are other problems with it as well (posted in a separate comment). Let's stick with a simple approach then (at least for now)

vladar (approved):

PR #232, opened 2022-11-01
Apollo devtools wrapper fallback fix

vejrj:


vejrj:
If window.__REMPL_APOLLO_DEVTOOLS_URL__ is not defined. We want to just let the devtools crash. We don't want to load subscriber from IIEFE local script.

vladar (commented):
LGTM

PR #241, opened 2022-11-19
Devtools HOTFIXES: empty cache, activity monitor recording and window opener support

vejrj:
If apollo-client's cache is empty and you switched from non-empty client. Then the empty client cache in UI is not updated

### DIFF
--- a/packages/rempl-apollo-devtools/src/publisher/publishers/apollo-cache-publisher.ts
+++ b/packages/rempl-apollo-devtools/src/publisher/publishers/apollo-cache-publisher.ts
@@ -31,23 +31,6 @@ export class ApolloCachePublisher {
     return client.cache.extract(true);
   }
 
-  private diffCaches(
-    currentCache: NormalizedCacheObject,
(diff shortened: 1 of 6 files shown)


packages/rempl-apollo-devtools/src/publisher/publishers/apollo-cache-publisher.ts
```diff
@@ -61,18 +44,17 @@ export class ApolloCachePublisher {
...
-    this.activeClient = activeClient;
-
     const serializedCacheObject = this.serializeCacheObject(activeClient);
 
     if (!serializedCacheObject) {
       return;
     }
 
-    if (!this.hasCacheChanged(this.lastCacheHistory, serializedCacheObject)) {
+    if (this.activeClient?.clientId !== activeClient.clientId) {
```
vladar, line 53: Can we compare instances? I believe clientId is tmp-specfic

```suggestion
    if (this.activeClient !== activeClient) {
```

vladar (approved):

PR #245, opened 2022-11-24
fix maglev styles

vejrj:


packages/rempl-apollo-devtools/apollo-devtools-maglev.css
```diff
@@ -0,0 +1,4 @@
+body > div[side="top"],
+body > div[side="right"] {
+  top: 40px;
```
alloy, line 3: Why not 42? 🤔

vejrj: Because it was just enough when i was testing it in maglev

PR #247, opened 2022-11-26
[TS-CODEGEN] models path fixed

vejrj:


packages/cli/src/supermassive.ts
```diff
@@ -70,8 +70,17 @@ function getContextPath(outputDir: string, contextImport: string | undefined) {
...
+    return contextImport;
+  }
+
   const contextDir = path.join(process.cwd(), contextImport);
-  return path.relative(outputDir, contextDir);
+
+  return path
+    .relative(outputDir, contextDir)
+    .split(path.sep)
+    .join(path.posix.sep);
```
freiksenet, line 83: It's annoying they don't provide a native api to do that :/

packages/ts-codegen/src/context/utilities.ts
```diff
@@ -23,3 +24,32 @@ export function createImportDeclaration(
     factory.createStringLiteral(from),
   );
 }
+
+function cleanRelativePath(relativePath: string) {
```
freiksenet, line 28: Nice.

packages/ts-codegen/src/context/__tests__/utitilites.test.ts
```diff
@@ -0,0 +1,56 @@
+import { getRelativePath } from "../utilities";
+import path from "path";
+
+describe("getRelativePath", () => {
+  test('"from" is package', () => {
```
freiksenet, line 5: Add a test for package like "foobar", so without a "@msteams"

freiksenet:
Looks good, minor nit.

freiksenet (approved):

PR #248, opened 2022-12-04
[CLI] removed unused identifiers and ts ignore

vejrj:


packages/cli/src/supermassive.ts
```diff
@@ -116,25 +124,37 @@ async function generateInterfaces(
...
+        path.join(outputDir, "models.interface.ts"),
+        PREPEND_TO_INTERFACES +
+          printer.printNode(
+            ts.EmitHint.SourceFile,
+            result.models,
+            result.models,
+          ),
+        { overwrite: true, scriptKind: ScriptKind.TS },
+      )
+      .fixUnusedIdentifiers()
```
alloy, line 138: Ok, so this is an API of ts-morph where it visits everything retroactively to fix it up? And how long did you say it takes compared to before?

Is there no option for us to pro-actively keep a "seen" list while emitting the body?

vejrj: It's ~3sec in total vs ~18 now. It may be fixed during the generation, but it will take some time. I just offer this as a temporary solution, which will allow us to get rid of typescript disable now.

alloy: What's the impact in TMP, how much time will this take when running for any dev? Otherwise, can we consider not removing the disabling of type-checking just yet in TMP, but have a test bed here in the repo where we disable it and only there invoke this `fixUnusedIdentifiers` method? (So that we can feel reasonable sure about things working when integrated into TMP without type-checking.)

vejrj: The time I wrote is in TMP. The check would help us now because every model has some unused fields.

alloy: I meant that we could still emit the comment in TMP until we've had the time to check-out a faster solution, but for now 18 sec also sounds fine.

alloy (approved):

PR #255, opened 2022-12-12
[TS-CODEGEN] remove unsused imports to be able to remove ts-morph

vejrj:


packages/cli/src/supermassive.ts
```diff
@@ -124,37 +116,28 @@ async function generateInterfaces(
...
+    await fs.writeFile(
+      path.join(outputDir, "models.interface.ts"),
+      PREPEND_TO_INTERFACES +
+        printer.printNode(ts.EmitHint.SourceFile, result.models, result.models),
+      { encoding: "utf-8" },
+    );
+
+    console.log(path.join(outputDir, "models.interface.ts"));
+
+    await fs.writeFile(
```
alloy, line 130: These two files can be written in parallel, right? i.e. `Promise.all(…)` them.

packages/ts-codegen/src/__tests__/index.test.ts
```diff
@@ -744,6 +744,310 @@ describe(generateTS, () => {
     });
   });
 
+  it("import an entity, which is used to implement interface and returned by resolver", () => {
```
alloy, line 747: Nit: the `it` is part of the sentence.

```suggestion
  it("imports an entity, which is used to implement interface and returned by resolver", () => {
```

packages/ts-codegen/src/__tests__/index.test.ts
```diff
@@ -744,6 +744,310 @@ describe(generateTS, () => {
...
+      extend type Query {
+        userById(id: ID!): Person
+      }
+    `);
+    expect(models).toMatchInlineSnapshot(`
+      "import type { EntityModel } from "@msteams/packages-test";
+      // Base type for all models. Enables automatic resolution of abstract GraphQL types (interfaces, unions)
+      export interface BaseModel {
+          __typename: string;
+      }
```
alloy, line 768: Nit: Why are we writing with 4 spaces instead of 2? I think we use 2 everywhere, so would be nice to do that here too.

packages/ts-codegen/src/__tests__/index.test.ts
```diff
@@ -744,6 +744,310 @@ describe(generateTS, () => {
     });
   });
 
+  it("import an entity, which is used to implement interface and returned by resolver", () => {
+    let { resolvers, models } = runGenerateTest(graphql`
+      extend schema @import(from: "@msteams/packages-test", defs: ["Entity"])
```
alloy, line 749: Hmmm, did we agree on shortening `definitions` in a spec? I'm not really a fan.

packages/ts-codegen/src/__tests__/index.test.ts
```diff
@@ -744,6 +744,310 @@ describe(generateTS, () => {
...
+
+      extend type Query {
+        userById(id: ID!): Person
+      }
+    `);
+    expect(models).toMatchInlineSnapshot(`
+      "import type { EntityModel } from "@msteams/packages-test";
+      // Base type for all models. Enables automatic resolution of abstract GraphQL types (interfaces, unions)
+      export interface BaseModel {
+          __typename: string;
```
alloy, line 767: Why is this not readonly?

packages/ts-codegen/src/__tests__/index.test.ts
```diff
@@ -744,6 +744,310 @@ describe(generateTS, () => {
...
+          readonly id: string;
+      }
+      "
+    `);
+    expect(resolvers).toMatchInlineSnapshot(`
+      "import type { PromiseOrValue } from "@graphitation/supermassive";
+      import type { ResolveInfo } from "@graphitation/supermassive";
+      import type { UserModel, PersonModel } from "./models.interface";
+      export declare namespace User {
+          export type id = (model: UserModel, args: {}, context: unknown, info: ResolveInfo) => PromiseOrValue<string>;
```
alloy, line 783: Mostly out of curiosity; I don't believe there is any point in exporting items from within the namespace, or am I missing something?

packages/ts-codegen/src/__tests__/index.test.ts
```diff
@@ -744,6 +744,310 @@ describe(generateTS, () => {
     });
   });
 
+  it("import an entity, which is used to implement interface and returned by resolver", () => {
+    let { resolvers, models } = runGenerateTest(graphql`
```
alloy, line 748: Out of curiosity, why do you use `let` instead of `const`?

packages/ts-codegen/src/context/index.ts
```diff
@@ -127,12 +140,20 @@ export class TsCodegenContext {
...
+        );
+        if (!filteredDefs.length) {
+          return;
+        }
+        return createImportDeclaration(
+          filteredDefs.map(({ typeName }) => addModelSuffix(typeName)),
+          from,
+        );
+      })
+      .filter(Boolean) as ts.ImportDeclaration[];
```
alloy, line 156: When you do an assertion, you loose benefits. I'm guessing it's the `map` that is loosing the precision? It would be better to pass this as a type parameter to map.

packages/ts-codegen/src/context/index.ts
```diff
@@ -127,12 +140,20 @@ export class TsCodegenContext {
...
 
   getAllImportDeclarations(): ts.ImportDeclaration[] {
-    return this.imports.map(({ defs, from }) =>
-      createImportDeclaration(
-        defs.map(({ typeName }) => addModelSuffix(typeName)),
-        from,
-      ),
-    );
+    return this.imports
+      .map(({ defs, from }) => {
```
alloy, line 144:
```suggestion
      .map<ts.ImportDeclaration>(({ defs, from }) => {
```

packages/ts-codegen/src/context/index.ts
alloy: Created example comment above, but unsure from the top of my head if that's the right parameter.

alloy (approved):
Some minor feedback, most notably the parallelising of file writing, otherwise LGTM 👍

packages/ts-codegen/src/__tests__/index.test.ts
vejrj: We need to be able to reach a field resolver within a type like `User.id` at the same time as `Conversation.id`. There was a conversation about this with you, I cannot quote, but you said something like that even though namespace is not recommended anymore and we should you module (based on the doc) this is the use-case which is an exception ?

packages/ts-codegen/src/__tests__/index.test.ts
vejrj: @freiksenet what do you think? I am fine with both.

packages/ts-codegen/src/__tests__/index.test.ts
vejrj: It can, it doesn't affect anything since it's implemented with `readonly` in the types, but I added it.

packages/ts-codegen/src/__tests__/index.test.ts
alloy: Yeah this was another topic. What I mean _here_ is that when you export a namespace, you don't also need to export the items within that exported namespace.

packages/ts-codegen/src/__tests__/index.test.ts
alloy: This doesn't need to block this PR btw.

packages/ts-codegen/src/__tests__/index.test.ts
```diff
@@ -744,9 +744,313 @@ describe(generateTS, () => {
...
+      }
+    `);
+    expect(models).toMatchInlineSnapshot(`
+      "import type { EntityModel } from "@msteams/packages-test";
+      // Base type for all models. Enables automatic resolution of abstract GraphQL types (interfaces, unions)
+      export interface BaseModel {
+          readonly __typename: string;
+      }
+      export interface PersonModel extends BaseModel, EntityModel {
+          __typename: string;
```
alloy, line 770: I assumed the readonly was applied the same to all these `__typename` cases, but seeing this is still here I guess not.

PR #259, opened 2023-01-24
Refactor codegen and add compat mod for legacy

freiksenet:


vejrj (approved):

packages/ts-codegen/src/__tests__/index.test.ts
```diff
@@ -489,15 +489,15 @@ describe(generateTS, () => {
             export type id = (model: UserModel, args: {}, context: unknown, info: ResolveInfo) => PromiseOrValue<string>;
         }
         export type PresenceInput = {
-            type: string;
+            readonly type: string;
```
alloy, line 492: These types are only used in resolvers and thus only ever read, right?

freiksenet: Correct.

PR #269, opened 2023-03-06
Add more legacy options to fix stuff for data-types-common

freiksenet:


packages/cli/src/supermassive.ts
```diff
@@ -45,6 +46,7 @@ export function supermassive(): Command {
     .option("-cn, --context-name [contextName]", "Context name")
     .option("-ei, --enums-import [enumsImport]", "from where to import enums")
     .option("-l, --legacy", "generate legacy types")
+    .option("-lm, --legacy-models", "do not use models for object types")
```
alloy, line 49: Nit: We don't need 1 letter aliases for options that should only be used in exceptional circumstances.

packages/ts-codegen/src/__tests__/index.test.ts
```diff
@@ -1975,6 +1975,55 @@ describe(generateTS, () => {
       "
     `);
   });
+
+  it("legacy no models", () => {
```
alloy, line 1979: It is no English sentence 🙃 

```suggestion
  it("does not generate models in legacy mode", () => {
```

packages/ts-codegen/src/__tests__/index.test.ts
```diff
@@ -1975,6 +1975,55 @@ describe(generateTS, () => {
...
+      export * from "./inputs.interface";
+      export declare namespace Bar {
+          export interface Resolvers {
+              readonly id?: id;
+          }
+          export type id = (model: Models.Bar, args: {}, context: unknown, info: ResolveInfo) => PromiseOrValue<string>;
+      }
+      "
+    `);
+  });
```
alloy, line 2026: It is not clear here what the expected output change is. Please keep tests lean so it clearly shows what the behaviour is. E.g. assuming that all other output is the same as normal codegen, can all this noise be reduced to something like this?

```ts
expect(models).not.toMatch("the model field/import/etc")
```

packages/ts-codegen/src/context/index.ts
```diff
@@ -118,6 +120,10 @@ export class TsCodegenContext {
     return this.options.legacyCompat;
   }
 
+  isLegacyNoObjectModels(): boolean {
```
alloy, line 123: ?

```suggestion
  shouldNotGenerateObjectModels(): boolean {
```

alloy (approved):
Functionality all looks good, just some style comments.

PR #275, opened 2023-03-08
Another small 4.9 deprecation fix

freiksenet:


alloy (approved):
Looks fine, but should this not be updating any tests? (Or rather, is there no coverage for this?)

PR #276, opened 2023-03-13
fix(graphql-codegen-typescript-typemap-plugin): correctly handle types with prefix and underscore

sjwilczynski:
Add a small fix to be able to generate type map for types with prefix ending with underscore like `OneGQL__`

### DIFF
--- a/packages/graphql-codegen-typescript-typemap-plugin/src/index.ts
+++ b/packages/graphql-codegen-typescript-typemap-plugin/src/index.ts
@@ -17,7 +17,10 @@ const graphqlCodegenTypeMapPlugin: PluginFunction = (
           `  "${typeName}": ${
             isScalarType(typesMap[typeName])
               ? `Scalars["${typeName}"]`
-              : pascalCase(typeName)
+              : typeName
+                  .split("_")
+                  .map((part) => pascalCase(part))
+                  .join("_")
           };`,
       ),
     "};\n",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3090,7 +3090,7 @@
   resolved "https://registry.yarnpkg.com/@tootallnate/once/-/once-2.0.0.tgz#f544a148d3ab35801c1f633a7441fd87c2e484bf"
   integrity sha512-XCuKFP5PS55gnMVu3dty8KPatLqUoy/ZYzDzAGCQ8JNFCkLXzmI7vNHCR+XpbZaMWQK/vQubr7PkYq8g470J/A==
 
-"@ts-morph/bootstrap@0.19.0":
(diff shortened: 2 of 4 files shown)


yarn.lock
```diff
@@ -3090,7 +3090,7 @@
   resolved "https://registry.yarnpkg.com/@tootallnate/once/-/once-2.0.0.tgz#f544a148d3ab35801c1f633a7441fd87c2e484bf"
   integrity sha512-XCuKFP5PS55gnMVu3dty8KPatLqUoy/ZYzDzAGCQ8JNFCkLXzmI7vNHCR+XpbZaMWQK/vQubr7PkYq8g470J/A==
 
-"@ts-morph/bootstrap@0.19.0":
+"@ts-morph/bootstrap@0.18.0":
```
alloy, line 3093: Interesting

packages/graphql-codegen-typescript-typemap-plugin/src/index.ts
```diff
@@ -17,7 +17,10 @@ const graphqlCodegenTypeMapPlugin: PluginFunction = (
           `  "${typeName}": ${
             isScalarType(typesMap[typeName])
               ? `Scalars["${typeName}"]`
-              : pascalCase(typeName)
+              : typeName
+                  .split("_")
+                  .map((part) => pascalCase(part))
+                  .join("_")
```
alloy, line 23: I think we should probably be using `@graphql-codegen` utility functions for the naming, but it was hard to figure out so I hadn't put in more work at the time. This seems fine for now 👍

yarn.lock
sjwilczynski: Hmmm got autogenerated but I can revert if needed. However, looking at the content, change seems reasonable :)

alloy (approved):
Add a change file and this is good to go

yarn.lock
alloy: Yeah I mean, if all tests pass then this should be moot.

packages/graphql-codegen-typescript-typemap-plugin/src/index.ts
sjwilczynski: Let me know exactly what you would want, maybe I will find time to do it properly. Which utility functions did have in mind? Any pointers?

alloy: Their "base visitor" has a bunch of utilities, but it wasn't clear to figure out which ones to use and how. Here's the public config docs https://the-guild.dev/graphql/codegen/docs/config-reference/naming-convention

PR #283, opened 2023-03-21
Add Grid view for the operations view

vejrj:
### Changes
- Added an Error boundary in the operations tracker, so that if there's a runtime exception, the whole apollo dev tools will not crash
- Added Grid view to show operations. This is a better UX that shows all the relevant info for an operation in a row.
- Clicking on a row in Grid, opens a detailed view of the operation.
- Implemented an `Affected Queries` view which shows which all operations caused a query to re-render
- Added Filter support in grid view to find the relevant operation faster and easier.
- Added readme for operations tracker.
- Updated `@fluent/react-components` package as only the latest package had the Grid view.
- Added resolution for `@fluentui/react-icons` as other packages were bringing the unstable version for this package.
- Added copying functionality to copy one or more operations.
- Grid view is virtualized. 
- Tested In Teams. 

image


packages/rempl-apollo-devtools/package.json
```diff
@@ -41,6 +43,9 @@
     "rempl": "^1.0.0-alpha.24",
     "uuid": "^8.3.2"
   },
+  "resolutions": {
```
jpsahoo14, line 46: This resolution can be removed.

PR #284, opened 2023-03-27
[SUPERMASSIVE] Initial execution hooks implementation

sergey-stoyan:
Initial implementation of the execution hooks:
 - beforeFieldResolve
 - afterFieldResolve
 - afterFieldComplete

Details:
 - Each hook accepts in arguments:
   - `ResolveInfo` of the field
   - Custom context passed into execution (same as passed into resolvers)
   - afterResolve/afterComplete also accept resolved value or error
 - Hooks aren't invoked for the default field resolvers
 - When error is thrown in the hook, it's returned in the response `errors` array

### DIFF
--- a/packages/supermassive/src/executeWithSchema.ts
+++ b/packages/supermassive/src/executeWithSchema.ts
@@ -18,6 +18,7 @@ export function executeWithSchema({
   operationName,
   fieldResolver,
   typeResolver,
+  hooks,
 }: ExecutionWithSchemaArgs): PromiseOrValue<ExecutionResult> {
   const schema = buildASTSchema(typeDefs);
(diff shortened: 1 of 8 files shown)


packages/supermassive/src/executeWithSchema.ts
```diff
@@ -18,6 +18,7 @@ export function executeWithSchema({
   operationName,
   fieldResolver,
   typeResolver,
+  hooks,
```
freiksenet, line 21: Let's rename to fieldExecutionHooks

sergey-stoyan: renamed

freiksenet (approved):
Looks good

PR #290, opened 2023-04-05
Documentation for learning how to GraphQL

alloy:


website/docs/learn-graphql/faq.md
```diff
@@ -0,0 +1,15 @@
...
+title: Frequently Asked Questions
+---
+
+# FAQ
+
+## When should I expose a relationship as an object type?
+
+Whenever you have a field called `somethingId` or `something_id`, it is most likely the case that you will want to expose `something` as a relationship with an object type.
+
+If you do need just the raw identifier, e.g. for passing to APIs _oustide_ of GraphQL, the user can query for `something.id`.
```
kerrynf, line 13:
```suggestion
If you do need just the raw identifier, e.g. for passing to APIs _outside_ of GraphQL, the user can query for `something.id`.
```

website/docs/learn-graphql/guides/resolution.md
```diff
@@ -0,0 +1,364 @@
...
+};
+```
+
+:::
+
+### 👍 Lazy resolution
+
+The second example is more flexible and efficient than the first one, where we can have explicit field resolvers for each field in the schema. These field resolver functions allow us to define how to derive the field's value from the data source.
+
+For example, if the `receivedAt` value would not already be formatted in the data source, we can define a resolver function for this field that calculates its human-readable value from the raw format. Here is how that field resolver function could look like:
```
kerrynf, line 169:
```suggestion
For example, if the `receivedAt` value would not already be formatted in the data source, we can define a resolver function for this field that calculates its human-readable value from the raw format. Here is what that field resolver function could look like:
```

website/docs/learn-graphql/guides/resolution.md
```diff
@@ -0,0 +1,364 @@
...
+        hour: "numeric",
+        hour12: true,
+        minute: "numeric",
+      });
+    },
+  },
+};
+```
+
+Similarly, the `participants` value in the data source is more likely to be a list of person IDs, than it is to be a list of full-fledged person objects. In this scenario, we need to issue an extra call to the data source to get the actual data. It should go without saying that we absolutely want this to be done only when the client needs this data, and not fetch it greedily in the `Query.conversations` root-field. Here is how that field resolver function could look like:
```
kerrynf, line 187:
```suggestion
Similarly, the `participants` value in the data source is more likely to be a list of person IDs, than it is to be a list of full-fledged person objects. In this scenario, we need to issue an extra call to the data source to get the actual data. It should go without saying that we absolutely want this to be done only when the client needs this data, and not fetch it greedily in the `Query.conversations` root-field. Here is what that field resolver function could look like:
```

website/docs/learn-graphql/guides/resolution.md
```diff
@@ -0,0 +1,364 @@
...
+      `${person.firstName} ${person.lastName}`,
+  },
+};
+```
+
+Here, the `person` argument has all the underlying data we need. We call such a source, **the model**. _Crucially_, the model type is **not** equal to the schema type. The model type is where the data comes _from_, the schema type's field is what the resolver transforms the data _to_.
+
+A model can be a raw response from the data source, an intermediate representation, or a full fledged model class instance. A raw data source response is the most basic form of a model. It could be a row from a database table, a document from a database, or a JSON object from an API response.
+
+An intermediate representation is a model that has some processing or transformation applied to it, perhaps ahead of time. For example, we might have a model that adds some computed properties during a background synchronization task. Note that this should **not** be transformation to full schema types,
```
kerrynf, line 279:
```suggestion
An intermediate representation is a model that has some processing or transformation applied to it, perhaps ahead of time. For example, we might have a model that adds some computed properties during a background synchronization task. Note that this should **not** be transformation to full schema types.
```

website/docs/learn-graphql/guides/resolution.md
```diff
@@ -0,0 +1,364 @@
...
+You will therefore absolutely want to pick a codegen tool that allows you to specify custom model typings for specific schema types. In the TypeScript space, such tools include:
+
+- Our own [graphitation supermassive codegen](https://github.com/microsoft/graphitation/tree/main/packages/cli), which allows you to annotate your schema definition with the model typings to use.
+- The popular [graphql-codegen](https://the-guild.dev/graphql/codegen) tool, which [allows you to provide configuration](https://the-guild.dev/blog/better-type-safety-for-resolvers-with-graphql-codegen) with schema type to model type mappings.
+
+:::
+
+## Performant data loading
+
+Integral to resolution of a graph of connected data, is that a query will end-up containing many entities of the same kind, or perhaps even contain the same entity multiple times. For instance, for each conversation fetch all participants—a classic N+1 problem.
```
kerrynf, line 304:
```suggestion
Integral to resolution of a graph of connected data, is that a query will end up containing many entities of the same kind, or perhaps even contain the same entity multiple times. For instance, for each conversation fetch all participants—a classic N+1 problem.
```

website/docs/learn-graphql/guides/resolution.md
```diff
@@ -0,0 +1,364 @@
...
+
+These entities might be necessary for unrelated parts of the application, but still, for performance reasons we want to be able to batch that entity data loading. [DataLoader](https://github.com/graphql/dataloader) is a utility used to abstract request batching in GraphQL. It allows you to reason about a batch of requests, **whithout** needing to do so in the field resolver functions—keeping them decoupled and without sacrificing the performance of batched data loading.
```
kerrynf, line 306:
```suggestion
These entities might be necessary for unrelated parts of the application, but still, for performance reasons we want to be able to batch that entity data loading. [DataLoader](https://github.com/graphql/dataloader) is a utility used to abstract request batching in GraphQL. It allows you to reason about a batch of requests, **without** needing to do so in the field resolver functions—keeping them decoupled and without sacrificing the performance of batched data loading.
```

website/docs/learn-graphql/guides/resolution.md
```diff
@@ -0,0 +1,364 @@
...
+
+#### Basic data loading
+
+Let’s look at how DataLoader could be used for the participants in our chat-list example. First we define the DatLoader instance, like so:
```
kerrynf, line 310:
```suggestion
Let’s look at how DataLoader could be used for the participants in our chat-list example. First we define the DataLoader instance, like so:
```

website/docs/learn-graphql/guides/thinking-in-graphql.md
```diff
@@ -0,0 +1,147 @@
...
+
+<table>
+<tr>
+<th>Conversation list</th>
+<th>UI components</th>
+<th>GraphQL query</th>
+</tr>
+<tr>
+<td>
+<img src={require("./images/SmallChatList.png").default} border="1" />
```
kerrynf, line 64: Image broken in preview here, not sure if there is an issue once it is in the 'real' view

website/docs/learn-graphql/guides/thinking-in-graphql.md
```diff
@@ -0,0 +1,147 @@
...
+type Conversation {
+  participantIDs: [ID]
+}
+```
+
+In this case, every root-field maps to a back-end service, and it of course does not return the full data for each related entity in its response payload, but rather contains foreign-keys to those related entities.
+
+Because we can only get the IDs of participants in a conversation, rather than the actual `Person` objects they refer to, we are being forced to make an aditional request for _each_ participant in all of the conversations in the list. This is the N+1 problem and forces the UI to perform a waterfall of requests. This in turn will lead to a slow loading experience or staggered UI rendering.
+
+TODO: Expand on the need for a single-source of truth, and refer to the not yet existant guide about how a graphql client works, combined with a FLUX store.
```
kerrynf, line 139: Tagging the TODO, not sure if that is to check in, or replace before hand..

website/docs/learn-graphql/intro.md
```diff
@@ -4,8 +4,28 @@ sidebar_position: 1
...
+
+The authors of this guide have observed that **the original premise of GraphQL is lost on most users**. On the one hand this speaks to the versatility of GraphQL, but on the other hand it means that missing key nuances can cause incorrect application of GraphQL, even in the exact same context it was designed for: **complex data-driven UI applications**.
+
+This guide aims to teach you everything you need to understand about GraphQL from that perspective, including how to design schemas, how to implement field resolvers, and how to effectively use this to build these data-driven UIs.
+
+This guide does not aim to replace [the canonical graphql.org site’s documentation](https://graphql.org/learn/). Some familiarity with GraphQL might be necessary for some sections, where possible the guide will link to the relevant existing documentation.
+
+:::caution
+Each section of this guide builds upon the knowledge gained in the previous section. Not taking this into account means you may end-up repeating the mistakes made in the community to learn about GraphQL outside of the context that it was designed for.
```
kerrynf, line 22:
```suggestion
Each section of this guide builds upon the knowledge gained in the previous section. Not taking this into account means you may end up repeating the mistakes made in the community to learn about GraphQL outside of the context that it was designed for.
```

website/docs/learn-graphql/guides/thinking-in-graphql.md
alloy: Works 👍 https://microsoft.github.io/graphitation/docs/learn-graphql/guides/thinking-in-graphql#-design-from-front-end-perspective

website/docs/learn-graphql/guides/the-design-of-graphql.md
```diff
@@ -41,7 +33,7 @@ The service sends some data down to the client, `ChatList` passes it on to its c
```
butaud, line 33: It seems like you're already assuming at this point that the top level component is the only one making a request to the backend service, since you talk about `ChatList` being the conduit for the data to get to its child components. Theoretically they could make their own queries (and I see this a lot in real code). It might be worth explaining this assumption or even working through why that is a bad pattern first.

website/docs/learn-graphql/guides/the-design-of-graphql.md
```diff
@@ -90,9 +80,9 @@ Because inherently, our rendering logic for `ConversationInfo` and its data-spec
 
 So if we want to do this, if we want each component to be able to specify its own data needs, how can we do so? The realization is that our data-specification has a key property that it needs to fulfill, which is composition.
 
-## Composition
+### Composition
 
-Composition in GraphQL is achieved by leveraging fragments, which are snippets of a query that can be composed together to form larger queries. These fragments are colocated with their components and composed into a tree that very much follows the shape of the component tree.
+Composition in GraphQL is achieved by leveraging fragments, which are snippets of a query that can be composed together to form larger queries. These fragments are co-located with their components and composed into a tree that very much follows the shape of the component tree.
```
butaud, line 85: I think a little more explanation here could help the reader key in on the important parts of the following code samples. Something like "Each component defines the specified properties it needs in a fragment, and then for any child components it references the fragment exported by that component. It knows that it has children with data dependencies, but it doesn't need to care about the details of that data."

website/docs/learn-graphql/guides/thinking-in-graphql.md
```diff
@@ -0,0 +1,147 @@
+---
```
butaud, line 1: I would argue that this page should actually come before the "Design of GraphQL" page. Its concepts are broader and don't depend on the fragment composition which is the focus of that page. But especially, this would resolve my concern on the other page that there is an unexplained assumption that you wouldn't be issuing queries from multiple levels of the component tree. This page explains why it's important for all of the data needs to roll up into one "broad query". If this page comes first, it will make more sense to work through how to keep the components loosely coupled even when making one broad query in the other page.

alloy: Oh, interesting thought. I see what you're saying. I guess the chronological order just follows so naturally to me, given the context I already have, but it would not be the first time that I have to learn that with different experiences (eg in TMP) what feels natural is the opposite.

Will try it out!

alloy: @butaud Ok, pushed. I'll let it bake for a bit; let me know what you think.

website/docs/learn-graphql/guides/the-design-of-graphql.md
alloy: With the order of the guides having changed, let's see how this reads.

website/docs/learn-graphql/guides/thinking-in-graphql.md
alloy: Added a new section on this: https://microsoft.github.io/graphitation/docs/learn-graphql/guides/thinking-in-graphql#generic-and-domain-specific

/cc @freiksenet Who gave me this TODO.

website/docs/learn-graphql/guides/the-design-of-graphql.md
alloy: Did done. Thanks!

alloy:
Alright, still lots of docs to write, but going to call this PR iteration done.

PR #303, opened 2023-05-07
[TS-CODEGEN] Hotfixes: empty files + unsused imports + @import directive fixed

vejrj:
Fixing the following stuff found in ts-codegen:
1. Doesn't create enum or input files when there are not inputs or enums defined in the schema
2. Doesn't define unused import statement
3. Fixing the @import directive (imports were missing in ts files)

### DIFF
--- a/packages/ts-codegen/src/context/index.ts
+++ b/packages/ts-codegen/src/context/index.ts
@@ -537,9 +537,7 @@ export function extractContext(
     },
     EnumTypeDefinition: {
       leave(node) {
-        if (!context.hasEnums) {
-          context.hasEnums = true;
-        }
+        context.hasEnums = true;
 
         context.addType({
           kind: "ENUM",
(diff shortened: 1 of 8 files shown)


alloy (commented):
Do the test changes cover all introduced code changes?

vejrj:
> Do the test changes cover all introduced code changes?

Yes, everything except the empty files was in the tests (wrong). Those undefined snapshots are there due to the empty files.

packages/ts-codegen/src/context/index.ts
```diff
@@ -543,6 +537,10 @@ export function extractContext(
     },
     EnumTypeDefinition: {
       leave(node) {
+        if (!context.hasEnums) {
```
alloy, line 540: Nit: Seems superfluous?

packages/ts-codegen/src/context/index.ts
```diff
@@ -579,6 +577,10 @@ export function extractContext(
     },
     InputObjectTypeDefinition: {
       leave(node) {
+        if (!context.hasInputs) {
```
alloy, line 580: Ditto

alloy (approved):

PR #304, opened 2023-05-10
[TS-CODEGEN] FIX: don't generate type extension in models

vejrj:


alloy (approved):
For my understanding, are these extensions lost in the ether or is the expectation that they would get generated from the type they extend _after_ first having been merged with the schema that has the base definition of the type?

The actual changes in the PR seem fine.

PR #308, opened 2023-05-25
Fix lack of support for default variables and values in supermassive

freiksenet:
Missing arguments without defaults were being ignored, now it's not the case.
- [x] Type annotator to add default values to all fields (so not just arguments, but missing ones too)
- [x] Check serialization and deserialization to see how it should actually be included in AST

### DIFF
--- a/packages/supermassive/src/__tests__/execute.test.ts
+++ b/packages/supermassive/src/__tests__/execute.test.ts
@@ -250,6 +250,38 @@ query Person($id: Int!) {
     }
     `,
   },
+  {
+    name: "Default value in variables",
+    document: `
+    query ($title: String! = "The Empire Strikes Back") {
+      searchFilmsByTitle(title: $title) {
+        title
+      }
+    }`,
+  },
+  {
+    name: "Default value",
+    document: `
+    {
+      searchFilmsByTitle {
+        title
+      }
+    }`,
+  },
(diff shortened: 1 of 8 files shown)


packages/supermassive/src/__tests__/execute.test.ts
```diff
@@ -319,5 +367,10 @@ async function compareResultsForExecuteWithoutSchema(
     schema,
     variableValues: variables,
   });
+  if (result.errors) {
+    for (const error of result.errors) {
+      console.error(error);
```
freiksenet, line 372: I've added this so that you can see stacktrace on failure.

packages/supermassive/src/ast/__tests__/addTypesToRequestDocument.test.ts
```diff
@@ -239,6 +324,19 @@ describe(addTypesToRequestDocument, () => {
           `,
         );
       }).toThrowError("Cannot find type for field: query.film.format");
+
+      expect(() => {
+        addTypesToRequestDocument(
+          schema,
+          graphql`
+            query {
+              film(ido: 42) {
```
vladar, line 333:
```suggestion
              film(id: 42) {
```

vladar (approved):

PR #309, opened 2023-05-29
[TS-CODEGEN] HOTFIX: Removed unused enums import in model files

vejrj:


packages/ts-codegen/src/models.ts
```diff
@@ -27,18 +27,20 @@ export function generateModels(
   const enumsStatements: ts.Statement[] = [];
 
   if (context.hasEnums) {
-    enumsStatements.push(
-      factory.createImportDeclaration(
-        undefined,
-        undefined,
-        factory.createImportClause(
-          false,
+    if (context.hasUsedEnumsInModels) {
```
vladar, line 30: Nit: can we merge two ifs into one?

vladar: Ignore me, I see now they are on different levels

vejrj: the first condition covers Imports and Exports. I want to export everything even when it's not used, nonetheless, I don't want to import it if it isn't used.

vladar (approved):

PR #310, opened 2023-06-01
Add defer and stream support to supermassive

freiksenet:


packages/apollo-mock-client/package.json
```diff
@@ -17,7 +17,7 @@
     "just": "monorepo-scripts"
   },
   "devDependencies": {
-    "@apollo/client": "^3.3.15",
+    "@apollo/client": "3.7.0",
```
alloy, line 20: Why are you doing this?

.azure-devops/graphitation-release.yml
```diff
@@ -1,6 +1,7 @@
 pr: none
 trigger:
   - main
+  - supermassive-v3
```
alloy, line 4: So I believe this will need to be merged into main

packages/supermassive/.eslintrc.json
```diff
@@ -1,4 +1,23 @@
...
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended"
+  ],
+  "rules": {
+    "no-unused-vars": "off",
+    "@typescript-eslint/no-unused-vars": [
+      "error",
+      { "argsIgnorePattern": "^_" }
+    ]
+  },
```
alloy, line 13: What's different here?

packages/supermassive/.eslintrc.json
```diff
@@ -1,4 +1,23 @@
...
+  "overrides": [
+    {
+      "files": ["**/__tests__/**"],
+      "rules": {
+        "@typescript-eslint/no-explicit-any": "off"
+      }
+    }
```
alloy, line 20: Perhaps we want this as a root option?

tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "paths": {
-      "@graphql-tools/utils": ["./node_modules/@graphql-tools/utils"]
+      "@graphql-tools/utils": ["./node_modules/@graphql-tools/utils"],
+      "graphql": ["./packages/supermassive/node_modules/graphql"]
```
alloy, line 13: So we're using latest version for typings for all monorepo packages, but not for actual runtime?

alloy (commented):
Nothing stands out, looks good thus far. Are there any specific supermassive differences in implementation that you could point out?

.azure-devops/graphitation-release.yml
freiksenet: Good point, thanks.

packages/apollo-mock-client/package.json
freiksenet: Your stuff breaks with apoll 3.7.15.

packages/supermassive/.eslintrc.json
freiksenet: That is a merge artifact, I did eslint there first. Will fix.

packages/supermassive/.eslintrc.json
freiksenet: Same as above.

tsconfig.json
freiksenet: Yeah, that was the only way I managed to get it to run.

packages/supermassive/src/__tests__/execute.test.ts
```diff
@@ -295,73 +313,505 @@ query Person($id: Int!) {
...
+      gender
+    }
+  }
+ `,
+    variables: {
+      limit: 5,
+      throwError: true,
+    },
+  },
+  // defer
```
vladar, line 360: There are a few cases where behavior is not quite clear from those tests:

1. Fields that exist inside both - normal selection set, and deferred one (especially nested ones):

```graphql
query {
  person {
    account { signupDate }
    ...PersonDetails @defer
   }
}

fragment PersonDetails on Person {
  account {
    __typename
    signupDate
  }
}
```
(same probably works with streams - especially deeply nested ones)

3. Include/skip inside defer/stream

4. Interplay between defer/stream and include/skip in various combinations, e.g.
```graphql
query($include: Boolean = false) {
  person {
    account { signupDate }
    ...PersonDetails @include(if: $include) @defer
   }
}

fragment PersonDetails on Person {
  account {
    __typename
    signupDate
  }
}
```

5. Error bubbling inside defer/stream

vladar: 5. `defer` for fragment on lists:

```graphql
query {
  someList {
     ...MyListItem @defer
  }
}
fragment MyListItem on ListItem {
  foo
}
```

6. `defer` for fragment on lists of union type:
```graphql
query {
  someListOnUnion {
     ...MyFoo @defer
     ...MyBar
  }
}
fragment MyFoo on UnionMemberFoo {
  foo
}
fragment MyBar on UnionMemberBar {
  bar
}
```

vladar: Different combinations of fragments on abstract types could be tricky (e.g. fragment on interface not deferred, fragment on specific - deferred and vice-versa

PR #318, opened 2023-06-19
supermassive-v3 alfa version branch added to release.yml

vejrj:


packages/supermassive/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@graphitation/supermassive",
   "license": "MIT",
-  "version": "2.4.8",
+  "version": "3.0.0-alpha.0",
```
freiksenet, line 4: Would this block any other supermassive releases? I think we should make it happen only in supermassive-v3 branch then.

PR #349, opened 2023-09-11
Add `types` conditional to package.json exports field

sverrejoh:
Every package has a `"types"` field, and an exports map `"exports"` field, but the `"types"` conditional is missing for each entry in the exports map. This breaks in TypeScript when using the `moduleResulution: "bundler"` options, as TypeScript then uses `exports` when available, but complains about missing types:

```
packages/nova-dev-app-providers/src/DevAppProvider.tsx:5:26 - error TS7016: Could not find a declaration file for module '@graphitation/apollo-react-relay-duct-tape'. 'node_modules/@graphitation/apollo-react-relay-duct-tape/lib/index.mjs' implicitly has an 'any' type.
  There are types at 'node_modules/@graphitation/apollo-react-relay-duct-tape/lib/index.d.ts', but this result could not be resolved when respecting package.json "exports". The '@graphitation/apollo-react-relay-duct-tape' library may need to update its package.json or typings.

5 import * as graphQL from "@graphitation/apollo-react-relay-duct-tape";
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1 error

```

This PR adds a types field for each package.


package-template/package.json
```diff
@@ -27,12 +27,13 @@
   "access": "public",
   "publishConfig": {
     "main": "./lib/index",
-    "types": "./lib/index.d.ts",
```
alloy, line 30: So this was in the wrong place?

alloy (approved):

PR #355, opened 2023-09-14
GraphQL 15 compatibility, test with GraphQL 15 by default

vladar:


alloy:
Yarn resolutions can be scoped, though, so you should be able to tell it to still install v17 for supermassive. iirc something like:

```
"packages/supermassive/graphql": "^17.0.0"
```

vladar:
@alloy sadly, this only works for "dependencies" and doesn't seem to work for "devDependencies" or "peerDependencies"

PR #359, opened 2023-09-26
Annotations for supermassive v3 ast

freiksenet:


packages/supermassive/src/schema/definition.ts
```diff
@@ -103,18 +107,67 @@ const enum InputValueKeys {
...
+  16: GraphQLDirectiveLocation.ENUM_VALUE,
+  17: GraphQLDirectiveLocation.INPUT_OBJECT,
+  18: GraphQLDirectiveLocation.INPUT_FIELD_DEFINITION,
+} as const;
+
 export type DirectiveName = string;
 export type DirectiveTuple = [
   name: DirectiveName,
-  arguments?: Record<string, unknown>, // JS values (cannot be a variable inside schema definition, so it is fine)
+  arguments: Record<string, unknown>, // JS values (cannot be a variable inside schema definition, so it is fine)
```
vladar, line 160: Directive usage should be OK without arguments, right?

```suggestion
  arguments?: Record<string, unknown>, // JS values (cannot be a variable inside schema definition, so it is fine)
```

packages/supermassive/src/schema/definition.ts
```diff
@@ -103,18 +107,67 @@ const enum InputValueKeys {
 export type InputValueDefinition = TypeReference | InputValueDefinitionTuple;
 export type InputValueDefinitionRecord = Record<string, InputValueDefinition>;
 
+const DirectiveLocation = {
+  QUERY: 0,
```
vladar, line 111: From my experience - using `0` could be a footgun :) One-indexed is probably safer? But up to you

vladar (approved):

packages/apollo-react-relay-duct-tape-compiler/package.json
```diff
@@ -41,7 +41,7 @@
   },
   "peerDependencies": {
     "graphql": "^15.0.0",
-    "@graphitation/supermassive-ast": "^2.0.1",
+    "@graphitation/supermassive": "^3.0.1",
```
alloy, line 44: @freiksenet @vladar This change makes it hard to ingest newer duct-tape-compiler builds into TMP, as we still install supermassive v2 in the root node_modules atm.

PR #361, opened 2023-09-27
Fix supermassive to decode directives and add tests to it

freiksenet:
Fix decode for directives

Add utilities to handle MVS documents


packages/supermassive/src/utilities/schemaFragmentFromMinimalViableSchemaDocument.ts
```diff
@@ -0,0 +1,25 @@
...
+): SchemaFragment {
+  const schemaDefinitions = createSchemaDefinitions(
+    document.definitions.map((def) => def.__defs),
+  );
+  const name =
+    operationName ||
+    document.definitions.find((def) => def.name?.value)?.name?.value ||
+    "(anonymous)";
+  return {
+    schemaId: name,
```
vladar, line 21: `schemaId` is needed to cache enum/scalar types per schema (I mean GraphQLEnumType and GraphQLScalarType instances) + for schema loader. If we pass `schemaId` as operation it won't really re-use those instances from the cache. So it was meant to be used with the actual schema id, i.e. "default" / "view schema" in our scenarios

packages/supermassive/src/utilities/__tests__/decodeASTSchema.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    const doc = cleanUpDocument(kitchenSinkSDL.document);
+    const encoded = [
+      mergeSchemaDefinitions(
+        { types: {}, directives: [] },
+        encodeASTSchema(doc),
+      ),
+    ];
+    const decoded = decodeASTSchema(encoded);
+    expect(encodeASTSchema(decoded)).toEqual(encoded);
+    expect(decoded).toMatchSnapshot();
```
vladar, line 29: Maybe `print(decoded)` instead? It is pretty hard to review those snapshots as is

PR #362, opened 2023-10-03
[DO NOT MERGE]: Repro error we have in our internal repo

kennetvu:
Initial load works fine; but if we trigger refetch which changes variables to main query, it breaks.
image

image

````
useCompiledFragment.ts:70  Uncaught TypeError: Cannot read properties of undefined (reading 'node')
    at useCompiledFragment (useCompiledFragment.ts:70:16)
    at useFragment (hooks.ts:123:31)
    at Todo (Todo.tsx:17:27)
    at WDYRFunctionalComponent (whyDidYouRender.js:1180:1)
    at renderWithHooks (react-dom.development.js:14985:1)
    at updateFunctionComponent (react-dom.development.js:17356:1)
    at beginWork (react-dom.development.js:19063:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:1)
    at invokeGuardedCallback (react-dom.development.js:4056:1)
```

Error is caused at this line:
https://github.com/microsoft/graphitation/blob/7c6aeac4065b17e017b99e94b5c8a41342747ec2/packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts#L69

![image](https://github.com/microsoft/graphitation/assets/1459265/6672afe5-c3c8-4f9a-9635-5aa512a9a4ca)


Data becomes undefined.

PR #373, opened 2023-10-26
[NOT MERGE] Repro and "fix" #362

dragoshomner:
This PR is a similar repro of the #362 issue and it shows how this problem can be fixed. However, it still should work with the directives + it still doesn't work in 1JS, so this issue might be related


alloy:
@dragoshomner Can you please provide some inline commentary about the changes and what/how they work?

dragoshomner:
> @dragoshomner Can you please provide some inline commentary about the changes and what/how they work?

@alloy I added a couple of commnets to make it more clear what I did. For short, I tried to simulate what we have in People App (PA). There, we a similar scenario, where we keep the variables passed to the query (sortBy and filterBy) in some React states, and we change their values when we receive an event (in this example is `refetch` function that sets the state). 

As I mentioned in the chat, when we change any state, the app crashes because data is undefined in `useFragment`. We had the same issue in the example app as well, but I was able to "fix" it by removing some directives (@argumentDefinitions, @includes) and setting `fist: 5` in `examples/apollo-watch-fragments/src/TodoList.tsx`. However, it's very fragile (as you see `@include` directive breaks it). I'm not able to fully understand what's happening, but this fragility increases exponentially in our app and it's impossible for me to figure out why. That's what we did in PA: https://office.visualstudio.com/Office/_git/1JS/pullrequest/2360178

The goal, as I mentioned during the Nova sync, is to migrate to duct-tape + pagination step-by-step, to reduce the risk. The first one is to replace the old nova compiler to duct-tape compiler in 1JS and make it work in TMP. I suppose that the old code should be compatible with both compilers (the only important change being that we removed the fragment interpolation). Then, the next step would be to use `useRefetchableFragment` and `usePaginationFragment` instead of the basic `useFragment`, but we are blocked on the first step. 

Hope it helps for more context :)

sjwilczynski:
@dragoshomner should we close this? Or was there something that hasn't been fixed yet?

PR #374, opened 2023-10-30
Embedded doc loader: speed this up to only apply work on relevant files

kenotron:
We bail a little early when "anyChanges" is not set as true. This means we skip the offset scanning work done - this speeds up the loader significantly (1 minute vs <1s over 20k files, for example, inside a sizeable repo).


kenotron:
Doing the 
BoingBoingBoingGIF
dance

alloy (approved):

PR #377, opened 2023-11-07
Fix unintentional breaking change in supermassive v3

vladar:
Supermassive V3 has introduced an unintentional breaking change by renaming `fieldNodes` property of `ResolveInfo` type to `fieldGroup`. 

This PR restores backwards compatibility with Supermassive V2 and graphql 16+

### DIFF
--- a/packages/supermassive/src/executeWithoutSchema.ts
+++ b/packages/supermassive/src/executeWithoutSchema.ts
@@ -764,8 +764,6 @@ export function buildResolveInfo(
     rootValue: exeContext.rootValue,
     operation: exeContext.operation,
     variableValues: exeContext.variableValues,
-
-    // Backwards-compat with graphql 16+ (17+ should use fieldGroup):
     fieldNodes: fieldGroup,
   };
 }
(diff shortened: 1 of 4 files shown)


sergey-stoyan (approved):

packages/supermassive/src/executeWithoutSchema.ts
```diff
@@ -764,6 +764,7 @@ export function buildResolveInfo(
     rootValue: exeContext.rootValue,
     operation: exeContext.operation,
     variableValues: exeContext.variableValues,
+    fieldNodes: fieldGroup,
```
freiksenet, line 767: Let's copy what tehy do in graphql-js, this isn't actually compatible it seems.

vladar: It is, graphql-js just did another round of changes in 17-alpha, so the code they have today in https://github.com/graphql/graphql-js/blob/8d7c8fccf5a9846a50785de04abda58a7eb13fc0/src/execution/execute.ts#L699 is not what we have in supermassive v3. The test in this PR confirms that it is compatible, resolving.

PR #382, opened 2023-12-07
Hacky relay transform

freiksenet:


packages/ts-transform-graphql-js-tag/src/index.ts
```diff
@@ -184,6 +197,109 @@ function getVisitor(
...
+  _sourceFile: ts.SourceFile,
+) {
+  const nodeVisitor: ts.Visitor = (node: ts.Node): ts.VisitResult<ts.Node> => {
+    if (ts.isObjectLiteralExpression(node)) {
+      if (
+        node.properties.findIndex((p) => {
+          console.log(
+            (p.name as any).text,
+            ((p as any).initializer as any).text,
+          );
```
alloy, line 212: Loggy

packages/ts-transform-graphql-js-tag/src/index.ts
```diff
@@ -184,6 +197,109 @@ function getVisitor(
...
+        }) !== -1
+      ) {
+        const params = node.properties.find(
+          (p) =>
+            ts.isPropertyAssignment(p) &&
+            ts.isStringLiteral(p.name) &&
+            p.name.text === "params",
+        );
+        if (params && ts.isPropertyAssignment(params)) {
+          console.log(params);
```
alloy, line 229: Ditto

packages/ts-transform-graphql-js-tag/src/index.ts
```diff
@@ -184,6 +197,109 @@ function getVisitor(
...
+          const paramsObject = params.initializer;
+          if (ts.isObjectLiteralExpression(paramsObject)) {
+            const text = paramsObject.properties.find(
+              (p) =>
+                ts.isPropertyAssignment(p) &&
+                ts.isStringLiteral(p.name) &&
+                p.name.text === "text",
+            );
+            console.log(text);
```
alloy, line 238: Ditto

packages/ts-transform-graphql-js-tag/src/index.ts
```diff
@@ -184,6 +197,109 @@ function getVisitor(
...
+                      p.name.text !== "text",
+                  )
+                  .concat([newText]),
+              );
+              const newParams = ts.factory.updatePropertyAssignment(
+                params,
+                params.name,
+                newParamsObject,
+              );
+              console.log(newParams);
```
alloy, line 270: Ditto

alloy (changes requested):
This is working in TMP.

One issue, however, is with [previously existing code](https://github.com/microsoft/graphitation/blob/1736b3a2c8ae316a495fdde91ebc626474f74945/packages/ts-transform-graphql-js-tag/src/index.ts#L348-L352). This hits when running in TMP. I get what the error is saying, but that's not what the conditional actually checks for. In the case of Relay this isn't actually ever an issue, so perhaps we shouldn't worry about it?

alloy (approved):
Alright, made those changes

PR #392, opened 2024-01-24
Generated documents can inline Maybe, Exact and basic Scalars

vejrj:


packages/graphql-codegen-typescript-operations/src/ts-operation-variables-to-object.ts
```diff
@@ -8,4 +9,11 @@ export class TypeScriptOperationVariablesToObject extends TSOperationVariablesTo
   ): string {
     return fieldType;
   }
+  protected getScalar(name: string): string {
```
freiksenet, line 12: we should check if we have those redefined - eg ID is often redefined, then it can't be used as a basic scalar.

freiksenet: I think also that if the scalar is defined as one of the "basic ts types", then it should be inlined too.

freiksenet (changes requested):

freiksenet (approved):

PR #397, opened 2024-01-29
Enums are inlined and used directly

vejrj:


packages/graphql-codegen-typescript-operations/src/visitor.ts
```diff
@@ -138,17 +145,27 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
         this.config.immutableTypes,
         this.config.namespacedImportName,
         enumsNames,
-        this.config.enumPrefix,
+        false,
```
vladar, line 148: Is this `false` because we do it separately in `PreResolveTypesProcessor` now? Should we keep the old `this.config.enumPrefix` value here when the original `CodegenPreResolveTypesProcessor` is used?

packages/graphql-codegen-typescript-operations/src/visitor.ts
```diff
@@ -138,17 +145,27 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
...
     this._declarationBlockConfig = {
       ignoreExport: this.config.noExport,
     };
   }
 
+  private getExportNames(): string[] {
+    const exportNames: string[] = Array.from(this.usedTypes);
+
+    if (!this.config.inlineCommonTypes) {
+      exportNames.push(...Array.from(this.usedEnums));
```
vladar, line 164: Nit: Array.from is redundant
```suggestion
      exportNames.push(...this.usedEnums);
```

packages/graphql-codegen-typescript-operations/src/ts-pre-resolve-types-processor.ts
```diff
@@ -0,0 +1,75 @@
...
+    return fields.map((field) => {
+      const fieldObj = schemaType.getFields()[field.fieldName];
+
+      const baseType = getBaseType(fieldObj.type);
+      let typeToUse = baseType.name;
+
+      const useInnerType = field.isConditional && isNonNullType(fieldObj.type);
+      const innerType = useInnerType
+        ? removeNonNullWrapper(fieldObj.type)
+        : undefined;
```
vladar, line 46: So if `fieldObj.type` is something like `[[[MyEnum!]!]!]!` - this will still work properly? Nested wrapping types are always tricky, 
let's add test cases for those (unless we already have those)?

E.g. 
1. All nullable: `[[[MyEnum]]]`
2. All non-nullable: `[[[MyEnum!]!]!]!`
3. A mix of those: `[[[MyEnum!]]!]`

vladar (approved):

PR #399, opened 2024-01-30
fix(apollo-react-relay-duct-tape): use correct id for compiler fragment, add check for partial result

sjwilczynski:
The aim of this PR is to fix an issue discovered while integrating refetchable fragments for Jana in TMP. In scope of this PR we:
* make sure that if there is an id variable on the query, it doesn't override the `id` used on the compiled fragment (reordering of spread and id in `useCompiledFragment`)
* add one more invariant, checking if `observableQuery` returned partial result (and failing in such case)
* add another invariant in case of data returned being null
* update readme about cache configuration

### DIFF
--- a/packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
+++ b/packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
@@ -5,6 +5,7 @@ import { useOverridenOrDefaultApolloClient } from "../../useOverridenOrDefaultAp
 
 import type { FragmentReference } from "./types";
 import type { CompiledArtefactModule } from "@graphitation/apollo-react-relay-duct-tape-compiler";
+import { Cache } from "@apollo/client/core";
 
 /**
  * @param documents Compiled watch query document that is used to setup a narrow
@@ -68,11 +69,15 @@ export function useCompiledFragment(
   if (result.partial) {
     invariant(
       false,
-      "useFragment(): Missing data expected to be seeded by the execution query document",
-      // TODO: queryInfo is private member and we can't access it, can we access last diff in other way?
-      // JSON.stringify(
-      //   observableQuery.queryInfo.lastDiff?.diff?.missing?.map((e) => e.path),
-      // ),
+      "useFragment(): Missing data expected to be seeded by the execution query document %s",
+      JSON.stringify(
+        // we need the cast because queryInfo and lastDiff are private but very useful for debugging
+        (
+          observableQuery as unknown as {
+            queryInfo: { lastDiff: { diff: Cache.DiffResult<unknown> } };
+          }
+        ).queryInfo.lastDiff?.diff?.missing?.map((e) => e.path),
+      ),
     );
   }
   let data = result.data;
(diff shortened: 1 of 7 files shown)


packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
```diff
@@ -65,6 +65,16 @@ export function useCompiledFragment(
...
 
   const result = observableQuery.getCurrentResult();
+  if (result.partial) {
+    invariant(
+      false,
+      "useFragment(): Missing data expected to be seeded by the execution query document",
+      // TODO: queryInfo is private member and we can't access it, can we access last diff in other way?
+      // JSON.stringify(
+      //   observableQuery.queryInfo.lastDiff?.diff?.missing?.map((e) => e.path),
+      // ),
```
alloy, line 75: I don’t believe so, maybe just cast to any and be really defensive? It should really aid in debugging

alloy (approved):

packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
sjwilczynski: Hmmm fair, do you know how we could trigger such scenario in tests just to make sure it works properly?

alloy: Probably disabling the fragment spreading fix for id would trigger it in people app.

packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
```diff
@@ -68,11 +69,15 @@ export function useCompiledFragment(
...
-      //   observableQuery.queryInfo.lastDiff?.diff?.missing?.map((e) => e.path),
-      // ),
+      "useFragment(): Missing data expected to be seeded by the execution query document %s",
+      JSON.stringify(
+        // we need the cast because queryInfo and lastDiff are private but very useful for debugging
+        (
+          observableQuery as unknown as {
+            queryInfo: { lastDiff: { diff: Cache.DiffResult<unknown> } };
+          }
+        ).queryInfo.lastDiff?.diff?.missing?.map((e) => e.path),
```
alloy, line 79: Should queryInfo also be treated as nullable?

sjwilczynski: Thanks! According to types it isn't but because we doing casts, I added null chaining now

packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
sjwilczynski: @alloy Ok, I can check try that tomorrow. However, I think it's a minor thing so if you agree let's merge the PR. If not I will try to figure out sth tomorrow

sjwilczynski: I didn't manage to trigger it on refetchable PR in TMP, maybe something changed about query/variables we have there. Hope we are ok merging like that and please reopen if you disagree

sjwilczynski:
Maybe we shouldn't merge it yet, I think I discovered an issue related the invariant locally

freiksenet (approved):

alloy (approved):

PR #405, opened 2024-02-28
Fragments containing @mask directive removed from the type

vejrj:


packages/graphql-codegen-typescript-operations/src/ts-selection-set-to-object.ts
```diff
@@ -0,0 +1,99 @@
...
+      Array<SelectionNode | string>
+    >();
+    const inlineFragmentSelections: InlineFragmentNode[] = [];
+    const fieldNodes: FieldNode[] = [];
+    const fragmentSpreads: FragmentSpreadNode[] = [];
+
+    for (const selection of selections) {
+      if (selection.directives?.some((value) => value.name.value === "mask")) {
+        break;
+      }
```
vladar, line 44: Mask should be only applicable to FRAGMENT_SPREAD. When seen on other selection kinds (FIELD, INLINE_FRAGMENT) we should throw instead.

packages/graphql-codegen-typescript-operations/src/ts-selection-set-to-object.ts
```diff
@@ -0,0 +1,109 @@
...
+      const hasMaskDirective = selection.directives?.some(
+        (value) => value.name.value === "mask",
+      );
+
+      if (hasMaskDirective) {
+        if (selection.kind === Kind.FRAGMENT_SPREAD) {
+          break;
+        } else {
+          throw new Error(
+            `Only Fragment spread can use @mask directive. Selection: ${selection}`,
```
freiksenet, line 51: There is error formatter in graphql-js, use it to print it nicely.

packages/graphql-codegen-typescript-operations/src/ts-selection-set-to-object.ts
```diff
@@ -0,0 +1,109 @@
...
+    const fragmentSpreads: FragmentSpreadNode[] = [];
+
+    for (const selection of selections) {
+      const hasMaskDirective = selection.directives?.some(
+        (value) => value.name.value === "mask",
+      );
+
+      if (hasMaskDirective) {
+        if (selection.kind === Kind.FRAGMENT_SPREAD) {
+          break;
```
freiksenet, line 48: Shouldn't that be a continue, not break?

freiksenet (changes requested):

freiksenet (approved):

PR #407, opened 2024-03-05
[SUPERMASSIVE] Execution hooks context

sergey-stoyan:
* Introduce custom execution hooks context.
* `afterFieldResolve` hook is now called right after resolver function returns; its `result` argument could now be a promise (for async resolvers).
* Export some `jsutils` functions, like `pathToArray`, `printPathArray` and `isPromise`.


alloy (approved):
Great, thanks!

PR #414, opened 2024-04-25
Add artifactDirectory option support to embedded-document-artefact-loader/ts-jest

mwangi-njuguna:
Adds support for explicitly passing `artifactDirectory` option in jest-config when using artifactDirectory in relay config.

### DIFF
--- a/change/@graphitation-embedded-document-artefact-loader-12b10da8-592d-4657-b0b2-50a40585c63d.json
+++ b/change/@graphitation-embedded-document-artefact-loader-12b10da8-592d-4657-b0b2-50a40585c63d.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add artifactDirectory option support to ts-jest",
+  "packageName": "@graphitation/embedded-document-artefact-loader",
+  "email": "mwanginjuguna59@gmail.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/embedded-document-artefact-loader/README.md
+++ b/packages/embedded-document-artefact-loader/README.md
@@ -63,6 +63,21 @@ const jestConfig = {
 };
 ```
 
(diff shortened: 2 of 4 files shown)


sjwilczynski (approved):

alloy:
@mwangi-njuguna Can you please elaborate on where/why you are using the `artifactDirectory` config? I understand what the config does in Relay, I contributed it, but I need to understand why we would need it inside MS?

Markionium:
@alloy the reason we're currently using it is because of the following:

- Some of the Nova components inside 1JS use different compilers (not all of them are on the Relay one yet). Which means that the generated files inside those packages are not Relay compatible.
- The other, probably primary, reason is to prevent writing files into other packages in the mono repo (This is a thing we try to stick to so the engineering system folks can optimise build times). 

e.g. If my package is `app-package` and I run the relay compiler there, I don't want `relay-compiler` to write files into a `dependency-package`. So what we've done so is to use `artifactDirectory` to generate all the files for an app in 1JS inside the apps package and have webpack pick up the files from there using either [@graphitation/embedded-document-artefact-loader/webpack](https://github.com/microsoft/graphitation/tree/main/packages/embedded-document-artefact-loader) or [@swc/plugin-relay](https://www.npmjs.com/package/@swc/plugin-relay).

This has been working great so far for the app, but right now we're trying to fix up our tests and thus the reason for this PR :)

As an FYI, this isn't for Jana, we're using this for Org Explorer.

Markionium (approved):

alloy:
I'm worried about runtime values being imported from a centralised artefact directory with a manual written import, which then wouldn't work in a host app that doesn't use a centralised artefact directory.. I guess it should normally only be typings, so I'll approve this, but let's be mindful of that.

alloy (approved):

PR #417, opened 2024-06-05
Add ban-directives rule

user1736:
Adds ban directives rule to the package. 
(also fixes typing issue in graphql-eslint plugin)


packages/graphql-eslint-rules/src/ban-directives.ts
```diff
@@ -0,0 +1,53 @@
...
+        },
+      },
+    },
+  },
+  create(context) {
+    const { bannedDirectives } = context.options[0];
+    return {
+      Directive(node) {
+        const directiveName = node.name?.value;
+        if (!directiveName || !node.loc) {
```
freiksenet, line 38: Why `!node.loc`?

user1736: fixed

freiksenet (approved):

PR #418, opened 2024-06-10
Revert afterFieldResolve hook invocation

sergey-stoyan:
- Revert back afterFieldResolve hook invocation way: it is called now when the field is actually resolved.
- Do not call execution hooks for `__typename`.
- Minor improvements in `printPathArray`: remove beginning `.` and use `+=` instead of map/join to speed up strings merging.

### DIFF
--- a/change/@graphitation-supermassive-9ce86754-6e03-40f7-b3aa-aff8ceddef1f.json
+++ b/change/@graphitation-supermassive-9ce86754-6e03-40f7-b3aa-aff8ceddef1f.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Revert back changes in execution where afterFieldResolve hook can return promise if async",
+  "packageName": "@graphitation/supermassive",
+  "email": "sergeystoyan@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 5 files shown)


vladar (approved):
LGTM

alloy:
> Revert back afterFieldResolve hook invocation way: it is called now when the field is actually resolved.

Hmm, afaik the executor refers to resolving and completing the value separately, so shouldn't this be called `afterFieldComplete` instead?

sergey-stoyan:
> > Revert back afterFieldResolve hook invocation way: it is called now when the field is actually resolved.
> 
> Hmm, afaik the executor refers to resolving and completing the value separately, so shouldn't this be called `afterFieldComplete` instead?

Yep, that's why we have 2 hooks: `afterFieldComplete` and `afterFieldResolve`. The first one is invoked after the field itself and all it's child fields (if field value is an object) are resolved, so that field resolution can be considered as `completed`. And the second -  when only field itself is resolved, and it's current implementation does not fully correspond to "afterResolve" concept, since the promise could be returned for async resolver. This PR changes that behavior to invoke afterResolve when the field value is actually resolved for both sync and async resolvers.

PR #419, opened 2024-06-11
add ban parameters rule

user1736:


packages/graphql-eslint-rules/src/__tests__/ban-parameters.test.ts
```diff
@@ -0,0 +1,53 @@
+import { GraphQLRuleTester } from "@graphql-eslint/eslint-plugin";
+import rule from "../ban-parameters";
+
+const ruleTester = new GraphQLRuleTester();
+
+ruleTester.runGraphQLTests("ban-parameters", rule, {
+  valid: [
+    {
+      code: `
+extend type Query {
```
user1736, line 10: would really appreciate more ideas for test cases

packages/graphql-eslint-rules/src/__tests__/ban-parameters.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  messageOfTheDay(date: Date!): String
+}`,
+      options: [
+        {
+          bannedParameters: [
+            {
+              keywords: ["force", "correlationId"],
```
alloy, line 17: I know that this rule doesn't yet ban any specific param, but it will be important to make sure that we can provide helpful reporting the dev that's specific to the banned parameter in question. E.g. for `correlationId` I expect to be able to print a message like: "This parameter is used as a side-effect to track correlated work through telemetry. This should be passed as context instead, see [URL] for details on how to do so with Apollo and Relay."

packages/graphql-eslint-rules/src/ban-parameters.ts
```diff
@@ -0,0 +1,99 @@
...
+      InputValueDefinition(node) {
+        for (const {
+          keywords,
+          description,
+          partialMatch,
+        } of bannedParameters) {
+          const paramName = node.name.value;
+          if (partialMatch) {
+            const matchingKeyword = keywords.find(
+              (keyword) => paramName.indexOf(keyword) !== -1,
```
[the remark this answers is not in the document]
user1736: The thought process was that we will have spelling enforced by lint rule, so there's no need to do case insensitive check 🤔  I can add it, if you think it's needed

packages/graphql-eslint-rules/src/__tests__/ban-parameters.test.ts
user1736: yes, config supports multiple groups so you can provide distinct descriptions for each group of keywords. I'll add test case that illustrates it.

packages/graphql-eslint-rules/src/ban-parameters.ts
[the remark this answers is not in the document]
alloy: I'm not sure I follow; what spelling? If we use `camelCase` for parameters, then either case can happen, no? E.g. `cacheBust` or `notCached`

vladar (approved):

PR #424, opened 2024-07-02
[SUPERMASSIVE] New execution hooks: beforeOperationExecute, beforeSubscriptionEventEmit and afterBuildResponse

sergey-stoyan:
Introduce 3 new execution hooks required to cover all possible cases related to execution telemetry implementation.

- beforeOperationExecute: is called when operation execution started (before it was impossible to detect operation execution start if operation selects several root fields).
- beforeSubscriptionEventEmit: is called before yielding a single event payload for subscription.
- afterBuildResponse: can be used to track operation execution completion.


freiksenet (approved):
Loo

alloy (approved):

PR #430, opened 2024-09-10
Fix refetch function to use new variables instead of old ones 

skipcheru:
When query variables change for the scenario where the variables are not set as defaults, the refetch function uses old variables instead of new variables and data is not fetched again. This is caused by the `fragmentReferenceWithOwnVariables` and `refetch` function missing as part of the useCallback dependencies. This PR adds those dependencies to fix it


sjwilczynski:
Let's make sure we have a unit test for our bug :)

sjwilczynski (approved):

PR #433, opened 2024-09-24
[SUPERMASSIVE] async before hooks added

vejrj:


packages/supermassive/src/hooks/types.ts
```diff
@@ -71,11 +71,13 @@ export interface AfterBuildResponseHook<ResolveContext = unknown> {
 }
 
 export interface BeforeOperationExecuteHook<ResolveContext = unknown> {
-  (args: BaseExecuteOperationHookArgs<ResolveContext>): void;
+  (args: BaseExecuteOperationHookArgs<ResolveContext>): void | Promise<void>;
```
sergey-stoyan, line 74: What about `BeforeFieldResolveHook`? Should we change return type for it as well?

packages/supermassive/src/__tests__/hooks.test.ts
sergey-stoyan, line 1: We also probably need a test to verify the `hookContext` is passed properly between hooks in a case when `BeforeFieldResolve` is async and `AfterFieldResolve` is sync

sergey-stoyan (approved):

PR #435, opened 2024-09-26
[WIP]: update to typescript 5.5

pavelglac:


pavelglac:
@microsoft-github-policy-service agree company="Microsoft"

packages/supermassive/src/executeWithoutSchema.ts
```diff
@@ -626,6 +626,10 @@ function executeSubscriptionImpl(
   const { groupedFieldSet } = collectFields(exeContext, rootTypeName);
 
   const firstRootField = groupedFieldSet.entries().next().value;
+  if (firstRootField === undefined) {
```
pavelglac, line 629: note: For ts 5.5 it is still `any` time but ts 5.6 has proper types and this type guard will be needed.

PR #436, opened 2024-09-27
bump typescript and lage

pavelglac:
Lage update: https://github.com/microsoft/lage/issues/733


packages/supermassive/src/executeWithoutSchema.ts
```diff
@@ -626,6 +626,10 @@ function executeSubscriptionImpl(
   const { groupedFieldSet } = collectFields(exeContext, rootTypeName);
 
   const firstRootField = groupedFieldSet.entries().next().value;
+  if (firstRootField === undefined) {
```
pavelglac, line 629: note: For ts 5.5 it is still any type but ts 5.6 has proper types and this type guard will be needed.

kenotron (approved):

packages/apollo-react-relay-duct-tape-compiler/package.json
```diff
@@ -38,13 +38,13 @@
...
     "ts-node": "^10.4.0",
-    "typescript": "^4.9.5",
+    "typescript": "^5.5.3",
     "@types/yargs": "^17.0.13"
   },
   "peerDependencies": {
     "graphql": "^15.0.0",
     "@graphitation/supermassive": "^3.5.2",
-    "typescript": "^4.3.5"
+    "typescript": "^5.0.0"
```
alloy, line 47: I assume this depends is here because we're using the compiler API. This does not follow semver, and so I think this might actually need to be higher in some cases. What do we use in TMP and how did you settle on 5.5.3 above?

alloy (approved):
Looks good overall, just wondering if the minimum ts version in peer deps needs to be bumped higher to ensure we actually get the tsc versions that we need.

packages/apollo-react-relay-duct-tape-compiler/package.json
alloy: (Same applies to all other package.json files ofc)

pavelglac: _5.5.X_ is the version that works in all package tests. The repo is compatible with _^5.6_, but some features of `ts-jest` are not, and those packages have strict versions. So, I wanted to have just one TS version for the whole monorepo.

Regarding the peerDep, sure, let's bump it to _^5.5.3_. TMP will satisfy it. I just wanted to have slightly better compatibility for folks who want to use it in different projects (and we are not using any specifics from 5.X releases). Changes made in: [bump peerDependencies](https://github.com/microsoft/graphitation/pull/436/commits/91c970708548e79b7c24e32043e3633e6d43a2c7)

alloy (approved):

PR #446, opened 2024-10-16
Upgrade web-dev-server on supermassive-todomvc

LSDima:
We have a High Vulnerability highlighted by our Component Governance check on the `webpack-dev-server` dependency. I'm upgrading the version of it to `^4.13.3` which is already used in the `graphitation` project.

### DIFF
--- a/examples/supermassive-todomvc/package.json
+++ b/examples/supermassive-todomvc/package.json
@@ -31,7 +31,6 @@
     "uuid": "^8.3.2",
     "webpack": "^5.76.0",
     "webpack-cli": "^4.6.0",
-    "webpack-dev-server": "^4.13.3",
     "zen-observable": "^0.8.15"
   },
   "resolutions": {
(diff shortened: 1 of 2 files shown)


examples/supermassive-todomvc/package.json
```diff
@@ -31,7 +31,7 @@
     "uuid": "^8.3.2",
     "webpack": "^5.76.0",
     "webpack-cli": "^4.6.0",
-    "webpack-dev-server": "^3.1.14",
+    "webpack-dev-server": "^4.13.3",
```
pavelglac, line 34: nitpick: I guess we can move it do devDependencies.

pavelglac (approved):

examples/supermassive-todomvc/package.json
LSDima: Thanks! I've moved it :)

Issue #331, opened 2023-07-14
Duct tape compiler not working with Typescript 5


sjwilczynski:
I am trying to integrate the new compiler in 1JS. During that process I stumbled upon error when trying to run the compiler. I was able to [reproduce this issue inside graphitation repo](https://github.com/microsoft/graphitation/commit/379fb1153cc11ec0b81eb1dd5ab0b9017d26d96a) by basically bumping TS to version 5 (which is what 1JS is using). The error is during generation is:

```
➜  apollo-watch-fragments (main) yarn duct-tape-compiler
yarn run v1.22.15
$ duct-tape-compiler --schema ./data/schema.graphql --src ./src --emitQueryDebugComments

Writing ts
ERROR:
Error writing modules:
Error: Debug Failure. Unhandled SyntaxKind: Unknown.
    at pipelineEmitWithHintWorker (C:\Users\stwilczy\Repos\graphitation\packages\apollo-react-relay-duct-tape-compiler\node_modules\typescript\lib\typescript.js:111514:13)  
    at pipelineEmitWithHint (C:\Users\stwilczy\Repos\graphitation\packages\apollo-react-relay-duct-tape-compiler\node_modules\typescript\lib\typescript.js:111058:9)
    at pipelineEmitWithComments (C:\Users\stwilczy\Repos\graphitation\packages\apollo-react-relay-duct-tape-compiler\node_modules\typescript\lib\typescript.js:114842:7)     
    at pipelineEmit (C:\Users\stwilczy\Repos\graphitation\packages\apollo-react-relay-duct-tape-compiler\node_modules\typescript\lib\typescript.js:111007:7)
    at emit (C:\Users\stwilczy\Repos\graphitation\packages\apollo-react-relay-duct-tape-compiler\node_modules\typescript\lib\typescript.js:110975:7)
    at emitTypeAliasDeclaration (C:\Users\stwilczy\Repos\graphitation\packages\apollo-react-relay-duct-tape-compiler\node_modules\typescript\lib\typescript.js:112937:7)     
    at pipelineEmitWithHintWorker (C:\Users\stwilczy\Repos\graphitation\packages\apollo-react-relay-duct-tape-compiler\node_modules\typescript\lib\typescript.js:111244:20)  
    at pipelineEmitWithHint (C:\Users\stwilczy\Repos\graphitation\packages\apollo-react-relay-duct-tape-compiler\node_modules\typescript\lib\typescript.js:111058:9)
[...]

PR #452, opened 2024-10-18
deps: bump all packages to react 18

sjwilczynski:
I wanted to make some changes in #450 but integrating `@types/relay-test-utils` requires `@types/react` version 18, so I thought - why not bump whole graphitation to v 18 of react.

In scope of this PR:
* we bump all react deps to 18
* we bring back <5 typescript requirement for duct tape compiler, check #331 (maybe we should have an UT for it?, for now I added it in build step of `apollo-watch-fragments`), cc @pavelglac 
* we fix `apollo-devtools-playground` app to start without errors

Change is verified with existing unit tests and by running start and build in examples and website


alloy:
I skimmed the PR quickly; you're still running the examples in legacy mode, right? That will be a necessity. What about tests, are/can those run in legacy mode?

sjwilczynski:
> I skimmed the PR quickly; you're still running the examples in legacy mode, right? That will be a necessity. What about tests, are/can those run in legacy mode?

I haven't updated the `render` to `createRoot` so examples are still using legacy mode. I also checked, tests and by default when using `react-test-renderer` (which is deprecated but we can postpone the migration until we move to concurrent mode) is using legacy root: 
image

alloy (approved):

PR #465, opened 2024-10-25
refactor(apollo-forest-run): replace const enums with ES modules

vladar:
Our monorepo builds sources using ESBuild, which doesn't support const enum inlining for typescript: https://github.com/evanw/esbuild/issues/128#issuecomment-1002341244

A quote from there:

> 4. Cross-module inlining only works when bundling

We don't bundle the code with ESBuild, just building ESM from TS, so const enums are not inlined in practice.

This PR implements an alternative approach by replacing const enums with wildcard ES module imports (with re-exports). By itself it doesn't change much, but when combined with minifers like Terser (in the actual bundler), inlining does occur.

### DIFF
--- a/packages/apollo-forest-run/src/diff/diffErrorKind.ts
+++ b/packages/apollo-forest-run/src/diff/diffErrorKind.ts
@@ -1,0 +1,3 @@
+export const MissingModelValue = 0,
+  MissingModelFields = 1,
+  MissingBaseFields = 2;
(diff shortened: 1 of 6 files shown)


packages/apollo-forest-run/src/diff/diffErrorKind.ts
```diff
@@ -0,0 +1,3 @@
+export const MissingModelValue = 0,
```
pavelglac, line 1: question: isnt fallback for enum object instead of desacralizing into single properties?

pavelglac (approved):

packages/apollo-forest-run/src/diff/diffErrorKind.ts
vladar: Literal object cannot be inlined because `const` keyword doesn't prevent object property mutations. For now, only scalar values with `const` keyword could be safely inlined by minifiers. 

`export * as Foo from "foo"` is not the same as an object with the same set of properties - bundlers like Webpack know that they can treat this object as immutable collection of ES exports and so can inline it too.

PR #476, opened 2024-11-13
fix(apollo-forest-run): do not leak variables when operation is evicted via LRU

vladar:


packages/apollo-forest-run/src/forest/types.ts
```diff
@@ -45,9 +46,9 @@ export type IndexedTree = {
 };
 
 export type IndexedForest = {
-  trees: MapLike<OperationDescriptor, IndexedTree>;
+  trees: MapLike<OperationId, IndexedTree>;
   extraRootIds: Map<NodeKey, TypeName>;
-  operationsByNodes: Map<NodeKey, Set<OperationDescriptor>>; // May contain false positives
+  operationsByNodes: Map<NodeKey, Set<OperationId>>; // May contain false positives
```
vladar, line 51: For reviewers: this is the essence of the change: since we do not clear `operationsByNodes` on LRU eviction - we don't want it to contain full descriptor. Cleanup of this map is out of scope for now - will address in future versions.

sergey-stoyan (approved):

PR #482, opened 2024-12-05
ForestRun: Explicit LRU eviction as a part of write transaction

vladar:
ForestRun evicts older trees automatically using LRU strategy. But the strategy before this PR was somewhat magical, using LRUMap for "trees" data. Basically any access to the via map `.get()` method caused the tree to be touched. This doesn't work well in certain cases. Plus, eviction happens in random moments and it is hard to control.

This PR exposes eviction as an explicit step in transaction as soon as forest size reaches a threshold for eviction. This way we have better control over it, and can adjust additional scenarios where we want to keep data in cache longer (e.g. with duct-tape integration)

### DIFF
--- a/packages/apollo-forest-run/src/ForestRun.ts
+++ b/packages/apollo-forest-run/src/ForestRun.ts
@@ -25,6 +25,7 @@ import {
   createOptimisticLayer,
   createStore,
   getEffectiveReadLayers,
+  maybeEvictOldData,
   removeOptimisticLayers,
   resetStore,
 } from "./cache/store";
@@ -481,6 +482,8 @@ export class ForestRun extends ApolloCache<any> {
         typeof optimistic === "string",
       );
     }
+    maybeEvictOldData(this.env, this.store);
+
     return result as T;
   }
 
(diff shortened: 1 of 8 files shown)


packages/apollo-forest-run/src/ForestRun.ts
```diff
@@ -481,6 +482,8 @@ export class ForestRun extends ApolloCache<any> {
         typeof optimistic === "string",
       );
     }
+    maybeEvictOldData(this.env, this.store);
```
freiksenet, line 485: Note that eg Relay does it scheduled during some idle time, do you reckon it's not worth here?

freiksenet (approved):

packages/apollo-forest-run/src/ForestRun.ts
vladar: In my local tests with a limit of 500 operations it takes ~2ms. So 2ms once per 500 cache writes - I decided not to add scheduling related complexity here. We can reconsider later of course.

PR #486, opened 2024-12-12
Add customScalars as parameter to the CLI

dragoshomner:
In our schema we could have custom scalars that are not covered in [TypeScriptTypeTransformers](https://github.com/microsoft/graphitation/blob/main/packages/apollo-react-relay-duct-tape-compiler/src/typescriptTransforms/TypeScriptTypeTransformers.ts#L84). For instance, we can have `DateTime` scalar that is not defined, but it should be mapped to `Date`. Now, since it is not found in those cases, it'd be mapped to `unknown`, as you can see in the following screenshot:

image

We added `DateTime` and we'll convert it on Typescript as a `string`, so that's how it's looking now:
image


alloy:
We discussed offline to add this to the existing string mapping, because it's a semantic type we have in 1GQL.

alloy (approved):

PR #487, opened 2024-12-13
Compiler: merging scalar and object error

pavelglac:
When we have union, one field is array of some type and the same field in different type is array of scalar we are throwing an error.

### DIFF
--- a/packages/apollo-react-relay-duct-tape-compiler/src/typescriptTransforms/TypeScriptGenerator.ts
+++ b/packages/apollo-react-relay-duct-tape-compiler/src/typescriptTransforms/TypeScriptGenerator.ts
@@ -121,6 +121,16 @@ function aggregateRuntimeImports(ast: ts.Statement[]) {
   }
 }
 
+function incompatibleTypesThrows(a: Selection, b: Selection) {
+  const isAScalar = !a.nodeSelections;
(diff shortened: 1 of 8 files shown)


patches/relay-compiler+12.0.0.patch
pavelglac, line 1: without this path I was not able to extend schema with union. The method for `parseUnionNode` was already implement but was missing in this switch statement.

freiksenet (approved):

packages/apollo-react-relay-duct-tape-compiler/src/typescriptTransforms/TypeScriptGenerator.ts
```diff
@@ -121,6 +121,16 @@ function aggregateRuntimeImports(ast: ts.Statement[]) {
   }
 }
 
+function incompatibleTypesThrows(a: Selection, b: Selection) {
+  const isAScalar = !a.nodeSelections;
+  const isBScalar = !b.nodeSelections;
+  if (isAScalar !== isBScalar) {
+    throw new Error(
+      `Type mismatch when merging selections. For union selection ensure all fragments quering for __typename`,
```
alloy, line 129: s/quering/query/

PR #498, opened 2025-01-17
feat(ts-codegen): option for default export

pavelglac:
## Changes

1. new `--generate-resolver-map` flag
2. Extracted options types for ts-codegen and remove duplications
3. CLI test for ts-codegen

## Output
When the flag is enabled we are generating default export for resolvers

```ts
export default interface ResolversMap {
    readonly User?: User.Resolvers;
    readonly Query?: Query.Resolvers;
}
```

note for `legacyEnumsCompatibility` the key was renamed https://github.com/microsoft/graphitation/pull/345 and therefore the test are not testing anything so it was removed in this PR.

### DIFF
--- a/packages/ts-codegen/README.md
+++ b/packages/ts-codegen/README.md
@@ -1,10 +1,46 @@
 # @graphitation/ts-codegen
 
-Package is used for [supermassive](../cli/src/supermassive.ts) cli to generate typescript code from a GraphQL document.
+This package generates Typescript definitions based on GraphQL Schema Definition Language.
 
-### Parameters
+## Parameters
 
-- `document` (DocumentNode): The GraphQL document to generate TypeScript code from.
-- `options` (object): An object containing various options for the code generation.
+- `document` (DocumentNode): The GraphQL document consisting the schema definitions to generate TypeScript code from.
+- `options` (GenerateTSOptions): An object containing various options for the code generation.
 
-TODO
+### Options
+
+Below is the list of all flags that can be passed to the code generator (see [codegen.ts](./src/codegen.ts)).
+
+```typescript
+export interface GenerateTSOptions {
+  outputPath: string;
+  documentPath: string;
+  contextTypePath?: string | null;
+  contextTypeName?: string;
+  enumsImport?: string | null;
+  legacyCompat?: boolean;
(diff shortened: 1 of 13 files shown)


freiksenet:
You need change files, let's do a major release because we are removing some APIs.

packages/ts-codegen/README.md
```diff
@@ -1,3 +1,10 @@
-# @graphitation/graphql-codegen
+# @graphitation/ts-codegen
+
+Package is used for [supermassive](../cli/src/supermassive.ts) cli to generate typescript code from a GraphQL document.
```
freiksenet, line 3:
```suggestion
This package generates Typescript definitions based on GraphQL Schema Definition Language.
```

packages/ts-codegen/README.md
```diff
@@ -1,3 +1,10 @@
...
+
+### Parameters
+
+- `document` (DocumentNode): The GraphQL document to generate TypeScript code from.
```
freiksenet, line 7:
```suggestion
- `document` (DocumentNode): The GraphQL document consisting the schema definitions to generate TypeScript code from.
```

packages/ts-codegen/README.md
```diff
@@ -1,3 +1,10 @@
...
+- `options` (object): An object containing various options for the code generation.
```
freiksenet, line 8: Could you add a codeblock with option type defs here?

freiksenet (commented):
LGTM, minor comments.

packages/ts-codegen/README.md
pavelglac: Sure, changed in https://github.com/microsoft/graphitation/pull/498/commits/94e2723505bde4b69a80721205de5e802a76ea22

packages/ts-codegen/README.md
pavelglac: Sure, changed in https://github.com/microsoft/graphitation/pull/498/commits/94e2723505bde4b69a80721205de5e802a76ea22

packages/ts-codegen/README.md
pavelglac: I just copy past the interface. I guess this is anyway used via CLI.

freiksenet (approved):

PR #502, opened 2025-01-23
Reverting back context bug fix in mapResultOrEventStreamOrPromise

vejrj:


packages/supermassive/src/__tests__/__snapshots__/execute.test.ts.snap
```diff
@@ -372,51 +372,6 @@ exports[`graphql-js snapshot check to ensure test stability subscription throw a
 }
 `;
 
-exports[`graphql-js snapshot check to ensure test stability subscription throw an error in the middle 1`] = `
```
freiksenet, line 375: Don't remove, comment out.

freiksenet (approved):

PR #508, opened 2025-01-31
feat: configurable auto-eviction

vladar:
Today ForestRun does automatic data eviction when the number of operation exceeds `maxOperationResult` option. This PR introduces an `autoEvict` option which allows disabling automated eviction and instead control when data eviction kicks in with `gc` method from Apollo cache interface.

Related options:

- `autoEvict`: default `true`
- `maxOperationCount`: default `1000`

### DIFF
--- a/packages/apollo-forest-run/src/__tests__/eviction.test.ts
+++ b/packages/apollo-forest-run/src/__tests__/eviction.test.ts
@@ -1,0 +1,145 @@
+import { ForestRun } from "../ForestRun";
+import { gql } from "./helpers/descriptor";
+
+it("evicts data automatically by default", () => {
+  const cache = new ForestRun({
+    maxOperationCount: 1,
+  });
+  const query = gql`
+    query ($i: Int) {
+      foo(i: $i)
+    }
+  `;
+
+  cache.write({ query, variables: { i: 0 }, result: { foo: 0 } });
+  cache.write({ query, variables: { i: 1 }, result: { foo: 1 } });
+  cache.write({ query, variables: { i: 2 }, result: { foo: 2 } });
+
+  const result0 = cache.read({ query, variables: { i: 0 }, optimistic: true });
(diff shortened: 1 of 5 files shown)


packages/apollo-forest-run/src/cache/env.ts
```diff
@@ -35,6 +34,9 @@ export function createCacheEnvironment(config?: CacheConfig): CacheEnv {
     },
     mergePolicies: new Map(),
     readPolicies: new Map(),
+    autoEvict: config?.autoEvict ?? true,
```
pavelglac, line 37: nitpick: Would be nice to have it sorted alphabetically someday

packages/apollo-forest-run/src/__tests__/eviction.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  expect(result0).toEqual(null);
+  expect(result1).toEqual({ foo: 1 });
+  expect(result2).toEqual({ foo: 2 });
+});
+
+it("allows to opt out operations from eviction", () => {
+  const cache = new ForestRun({
+    maxOperationCount: 1,
+    autoEvict: false,
+    nonEvictableQueries: new Set(["b"]), // root-level field
```
pavelglac, line 114: question: maxOperationCount is actually increased by length of nonEvictableQueries, right?

pavelglac (approved):

packages/apollo-forest-run/src/__tests__/eviction.test.ts
vladar: yes, by design

packages/apollo-forest-run/src/cache/env.ts
vladar: Maybe someday :) I prefer having related options grouped together - so if we sort alphabetically we will need to rename them to have a common prefix for the same "group" (or introduce another level of objects for grouping). Which is a breaking change.

PR #511, opened 2025-02-11
Make root fields mandatory

user1736:
Makes properties in `Resolvers` interface for query/mutation/subscription mandatory.
Also ensures that `Query`/`Mutation`/`Subscription` property in `ResolverMap` is mandatory.

### DIFF
--- a/packages/cli/src/__tests__/typeDefsToImplicitResolvers.test.ts
+++ b/packages/cli/src/__tests__/typeDefsToImplicitResolvers.test.ts
@@ -52,4 +52,25 @@ describe(supermassive, () => {
       ).toMatchSnapshot();
     }
   });
+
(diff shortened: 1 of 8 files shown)


pavelglac (approved):

packages/cli/src/__tests__/typeDefsToImplicitResolvers.test.ts
```diff
@@ -52,4 +52,25 @@ describe(supermassive, () => {
       ).toMatchSnapshot();
     }
   });
+
+  it("should generate interfaces with --mandatory-root-operation-types --generate-resolver-map", async () => {
```
freiksenet, line 56: Why do we have it in this file :D 

Let's just rename it to supermassive.test.ts?

user1736:
Abandoning in favour of #512

PR #512, opened 2025-02-12
Make extended types mandatory

user1736:


packages/ts-codegen/src/resolvers.ts
```diff
@@ -29,10 +29,17 @@ const getResolverTypes = (context: TsCodegenContext): ResolverType[] => {
...
+  mandatoryResolverTypes: boolean;
+}
+
 export function generateResolvers(
   context: TsCodegenContext,
-  generateResolverMap?: boolean,
+  options: Options,
 ): ts.SourceFile {
+  const { generateResolverMap, mandatoryResolverTypes } = options;
+  console.log(`DEBUG:${mandatoryResolverTypes}`);
```
freiksenet, line 42: [nit] debug console

user1736: fixed

PR #517, opened 2025-02-17
Replace useFragment with Apollo's native

fidelman:
- replace custom duct-tape `useFragment` implementation with Apollo's native `useFragment` https://www.apollographql.com/docs/react/data/fragments#usefragment
- dep-up @apollo/client to 3.8.1 as `useFragment` is stable there

### DIFF
--- a/patches/@apollo+client+3.6.10.patch
+++ b/patches/@apollo+client+3.6.10.patch
@@ -106,6 +106,57 @@ index a2c1b93..d28dddf 100644
          fieldName: fieldName,
          storeFieldName: storeFieldName,
          variables: variables,
+diff --git a/node_modules/@apollo/client/react/hooks/hooks.cjs b/node_modules/@apollo/client/react/hooks/hooks.cjs
+index 3a2234a..d7b5029 100644
+--- a/node_modules/@apollo/client/react/hooks/hooks.cjs
++++ b/node_modules/@apollo/client/react/hooks/hooks.cjs
+@@ -616,7 +616,46 @@ function useReactiveVar(rv) {
+     return value;
+ }
+ 
++function useFragment(options) {
++    var cache = useApolloClient().cache;
++    var fragment = options.fragment, fragmentName = options.fragmentName, from = options.from, _a = options.optimistic, optimistic = _a === void 0 ? true : _a, rest = tslib.__rest(options, ["fragment", "fragmentName", "from", "optimistic"]);
++    var diffOptions = tslib.__assign(tslib.__assign({}, rest), { returnPartialData: true, id: typeof from === "string" ? from : cache.identify(from), query: cache["getFragmentDoc"](fragment, fragmentName), optimistic: optimistic });
++    var resultRef = React__namespace.useRef();
++    var latestDiff = cache.diff(diffOptions);
++    var getSnapshot = function () {
++        var latestDiffToResult = diffToResult(latestDiff);
++        return resultRef.current &&
++            equality.equal(resultRef.current.data, latestDiffToResult.data)
++            ? resultRef.current
++            : (resultRef.current = latestDiffToResult);
++    };
++    return useSyncExternalStore(function (forceUpdate) {
++        var lastTimeout = 0;
++        var unsubcribe = cache.watch(tslib.__assign(tslib.__assign({}, diffOptions), { immediate: true, callback: function (diff) {
++                if (!equality.equal(diff, latestDiff)) {
++                    resultRef.current = diffToResult((latestDiff = diff));
++                    lastTimeout = setTimeout(forceUpdate);
++                }
++            } }));
++        return function () {
++            unsubcribe();
++            clearTimeout(lastTimeout);
++        };
++    }, getSnapshot, getSnapshot);
++}
++function diffToResult(diff) {
++    var result = {
++        data: diff.result,
++        complete: !!diff.complete,
++    };
++    if (diff.missing) {
++        result.missing = utilities.mergeDeepArray(diff.missing.map(function (error) { return error.missing; }));
++    }
++    return result;
++}
++
+ exports.useApolloClient = useApolloClient;
++exports.useFragment = useFragment;
+ exports.useLazyQuery = useLazyQuery;
+ exports.useMutation = useMutation;
+ exports.useQuery = useQuery;
 diff --git a/node_modules/@apollo/client/react/hooks/index.d.ts b/node_modules/@apollo/client/react/hooks/index.d.ts
 index a590efb..458f53f 100644
 --- a/node_modules/@apollo/client/react/hooks/index.d.ts


fidelman:
converting it to Draft, as TS is failing due to apollo/client dep-up. Resolving...

examples/apollo-watch-fragments/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@apollo/client": "~3.6.0",
+    "@apollo/client": "3.8.1",
```
alloy, line 6: Is this the version we are using in TMP today? If not, I believe we agreed to use TMP's version and include a patch-package.

packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
```diff
@@ -28,73 +29,102 @@ export function useCompiledFragment(
...
+    "useFragment(): Expected the fragment reference to have a string id. " +
+      "Did you forget to invoke the compiler?",
+  );
+  invariant(
+    metadata,
+    "useFragment(): Expected metadata to have been extracted from " +
+      "the fragment. Did you forget to invoke the compiler?",
+  );
 
-  const client = useOverridenOrDefaultApolloClient();
```
alloy, line 32: This is missing from the new implementation, and I do recall that I added this for a reason. It would be good to check the history for when I added this ability and see if there's a reason left in those commits.

alloy:
Great to see how simple this turned out to be! With this in place, and especially if we end up not needing to use TMP's patched Apollo version, I think we could actually go out and advertise duct-tape as something others could use. (But we'd need to discuss that more first.)

packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
```diff
@@ -28,73 +29,102 @@ export function useCompiledFragment(
     "useFragment(): Expected a `watchQueryDocument` to have been " +
       "extracted. Did you forget to invoke the compiler?",
   );
+  const from = fragmentReference.id;
+  invariant(
+    typeof from === "string",
+    "useFragment(): Expected the fragment reference to have a string id. " +
+      "Did you forget to invoke the compiler?",
+  );
```
vladar, line 37: I don't think we can do it here, because `fragmentReference.id` can be `undefined`, which is expected according to comments here:
 
image

packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
```diff
@@ -28,73 +29,102 @@ export function useCompiledFragment(
...
-      JSON.stringify(
-        // we need the cast because queryInfo and lastDiff are private but very useful for debugging
-        (
-          observableQuery as unknown as {
-            queryInfo?: { lastDiff?: { diff?: Cache.DiffResult<unknown> } };
-          }
-        ).queryInfo?.lastDiff?.diff?.missing?.map((e) => e.path),
-      ),
-    );
-  }
```
vladar, line 82: We must keep this invariant, but trigger it if `useFragment` returns `!result.complete`

examples/apollo-watch-fragments/package.json
fidelman: nope, TMP uses 3.6.10. I gonna leave the graphitation as it is and patch it

fidelman:
> Great to see how simple this turned out to be! With this in place, and especially if we end up not needing to use TMP's patched Apollo version, I think we could actually go out and advertise duct-tape as something others could use. (But we'd need to discuss that more first.)

sure it is!

examples/apollo-watch-fragments/package.json
fidelman: because dep-up duct-tape's Apollo will cause problem in TMP as it is patched there. Discussed w/@vladar so far the easiest way is to patch duct-tape's Apollo (btw it is already it is), and only dep-up duct-tape in TMP.

packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
fidelman: `fragmentReference.id` is required for useFragment args to connect fragment with a query. @alloy do you think we can accept `fragmentReference.id` is nillable? Previous implementation did not have such invariant, but cache.watch did not require it either

packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
fidelman: thanks, fixed

packages/apollo-react-relay-duct-tape/src/storeObservation/compiledHooks/useCompiledFragment.ts
fidelman: @alloy TLDR; we do not need that in the new implementation

### My explanation
The old implementation were introduced by 3 commits:
1. https://github.com/microsoft/graphitation/commit/bf2f03eaf109113b7e3bbff4c12b9947a0f7e784 - replace `useState` with `observableQuery.getCurrentResult()`, commit text: *Only trigger useState change when we really want it to*
2. https://github.com/microsoft/graphitation/commit/21f61ee38e8682164f23b2a06559e504918676d8#diff-72832f4a1f9e169a840a8d60938b0a66249d0dc384d1cf3124f6ca049747f63f - replace forceUpdate with a dedicated hook, commit text: *[apollo-react-relay-duct-tape] Implement useRefetchableFragment*
3. https://github.com/microsoft/graphitation/commit/87cb259cf55adbca0d1154d578aa4aca9d95593c, added a comment  **// Unclear why, but this yields twice with the same results, so skip one.**, commit text: *[ARRDT] Cleanup*

My assumption, correct me if you recall, The subscription was triggered twice (for unknown reason), "skip" was to introduce to skip the first trigger and then `forceUpdate` triggers the hook again getting an actual value from the `observableQuery.getCurrentResult();`.

Here is how it looks without "skip":

Screenshot 2025-02-24 173550

Here is how it looks w/ "skip" aka "old version":

Screenshot 2025-02-24 173609
Screenshot 2025-02-24 173836.

And here is the logs of the new version:
Screenshot 2025-02-24 174213
As you can see it is identical to the old version, this is why I think we are good to remove that.

yarn.lock
```diff
@@ -90,12 +90,13 @@
   dependencies:
     "@babel/highlight" "^7.18.6"
 
-"@babel/code-frame@^7.25.7":
```
vladar, line 93: I assume we can revert all changes to yarn.lock now that we have a patch?

PR #518, opened 2025-02-18
feat(ts-codegen): translate @import into namespace imports

user1736:
[WIP]


freiksenet:
You should check if that needs a config file or it might break TMP. ALso all manually committed files need to be updated.

freiksenet (approved):

freiksenet (approved):

PR #520, opened 2025-02-24
fix(apollo-forest-run): do not visit the same chunk twice when reading

vladar:
This PR fixes a quirk of ForestRun that was known before, but proved to be really detrimental for perf in some edge cases in the real world app.

In short, embedded objects (i.e. objects without ids) could be wastefully re-visited many times while reading object value in presence of missing fields (which is common with duct-tape and some other scenarios).

Here we simply track which objects were visited during read and don't re-visit them again.

### DIFF
--- a/packages/apollo-forest-run/src/values/execute.ts
+++ b/packages/apollo-forest-run/src/values/execute.ts
@@ -64,6 +64,7 @@ type Context<TModel, TListModel> = {
   operation: OperationDescriptor;
   resolver: Resolver<TModel, TListModel>;
   incompleteValues: IncompleteValues;
+  visited: Map<SourceObject | SourceCompositeList, Set<TModel | TListModel>>;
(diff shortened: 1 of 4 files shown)


sergey-stoyan (approved):

hanys7 (approved):

packages/apollo-forest-run/src/values/execute.ts
```diff
@@ -300,14 +308,16 @@ function executeCompositeListSelection<TModel, TListModel>(
 
   let incompleteItems: Set<number> | undefined;
   if (draft) {
-    incompleteItems = incompleteValues.get(draft);
+    incompleteItems = incompleteValues.get(draft) as Set<number> | undefined;
```
pavelglac, line 311: nitpick: i guess you do not need to overcast it. It already returns the same type.

PR #528, opened 2025-03-20
fix(duct-tape): do not unsubscribe in loading state

pavelglac:
We are unsubscribing only if the state is not loading. In case of store-and-network we get the data immediately and then unsubscribe. With this change we returns the data and waits until the request finish.


vladar (approved):
Good catch 👍

PR #536, opened 2025-03-30
Change build target to ES2018; inline asserts

vladar:
## ES2018

`ES2018` is the earliest version which supports object spread syntax, i.e. `{ ...cloned }`. With our current build target (`es6`) we rely on TS polyfill for it which is significantly slower. For perf-sensitive packages like apollo-forest-run it is important.

## Inline asserts

This takes inspiration from a [similar approach](https://github.com/graphql/graphql-js/blob/0a848cc4c781fb79b2e55eb8d011675dd612524f/resources/inline-invariant.js#L4) taken in graphql-js. In our case it gives us two things:

1. Perf savings due to inlining (`assert` function is not called)
2. Additional debugging info for minified version, or version without access to to stack trace

### DIFF
--- a/scripts/just.config.ts
+++ b/scripts/just.config.ts
@@ -10,6 +10,8 @@ import {
 import * as path from "path";
 import * as glob from "fast-glob";
 import * as fs from "fs";
+import { transformFileSync } from "@babel/core";
+import { inlineAssert } from "./inline-assert";
 
 export const types = () => {
   return tscTask({
@@ -22,7 +24,7 @@ export const build = () => {
   const baseEsbuildOptions: EsbuildBuildOptions = {
     entryPoints: glob.sync(["src/**/*.{ts,tsx}", "!src/**/__tests__/**"]),
     outdir: "lib",
-    target: "es6",
+    target: "es2018",
     sourcemap: "external",
   };
   return parallel(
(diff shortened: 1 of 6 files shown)


scripts/just.config.ts
```diff
@@ -32,6 +34,24 @@ export const build = () => {
       bundle: true,
       outExtension: { ".js": ".mjs" },
       plugins: [
+        {
+          name: "inline-assert",
+          setup(build) {
+            // TODO: switch to SWC or tree-sitter
+            build.onLoad({ filter: /apollo-forest-run/ }, async (args) => {
```
pavelglac, line 41: question: why we filter it? Would be nice to have one building config for all. Also why not to also grep **devAssert** from `graphql-js`?

scripts/just.config.ts
```diff
@@ -22,7 +24,7 @@ export const build = () => {
   const baseEsbuildOptions: EsbuildBuildOptions = {
     entryPoints: glob.sync(["src/**/*.{ts,tsx}", "!src/**/__tests__/**"]),
     outdir: "lib",
-    target: "es6",
+    target: "es2018",
```
pavelglac, line 27: nitpick: shouldn't we produce change files all packages?

pavelglac (approved):

scripts/just.config.ts
vladar: 1. Babel transforms are expensive at build-time. Without the filter it will be applied to every package and we don't want that. 
2. I need this for ForestRun, but other packages should be opted-in as needed (and patterns adjusted accordingly). So far I don't know any other packages in our repo, needing this.

scripts/just.config.ts
vladar: I couldn't find a quick way to bump all packages with beachball. Do you know one?

pavelglac: you can try `yarn beachball change --all`

vladar: Great, thanks! I'll do it in the follow up.

PR #541, opened 2025-04-02
Unify codegen exports

user1736:
With this change codegen won't emit `inputs.interface.ts` and instead put all input types into `models.interface.ts`. This change allows to use input types with @import directive.


packages/ts-codegen/src/__tests__/index.test.ts
```diff
@@ -527,14 +515,15 @@ describe(generateTS, () => {
...
+            readonly name?: string | null;
+        };
         "
       `);
       expect(resolvers).toMatchInlineSnapshot(`
         "import type { PromiseOrValue } from "@graphitation/supermassive";
         import type { ResolveInfo } from "@graphitation/supermassive";
         import * as Models from "./models.interface";
-        import * as Inputs from "./inputs.interface";
-        export * from "./inputs.interface";
```
user1736, line 537: I wonder if these exports were used. Should I `export * from "./models.interface"` to cover for it? 🤔

user1736: skimmed through TMP codebase and there doesn't seem to be usages for this particular export. There's only a few instance of direct imports from `inputs.interface` but those look easy enough to fix.

If there's no issues with the change itself, I can prepare a patch to test if TMP CI breaks at the very least.

user1736: Tested latest changes with TMP. Added tests for missing cases.

freiksenet (approved):

PR #543, opened 2025-04-03
feat(website): bump docusaurus

pavelglac:
mermaid 10.0.3 fix the security issue, this PR bumping the version to the safe one. https://github.com/cure53/DOMPurify/security/advisories/GHSA-mmhx-hmjr-r674
image

### DIFF
--- a/examples/supermassive-todomvc/package.json
+++ b/examples/supermassive-todomvc/package.json
@@ -3,7 +3,7 @@
   "license": "MIT",
   "main": "index.js",
   "private": true,
-  "version": "1.0.1-0",
+  "version": "1.0.1-1",
   "scripts": {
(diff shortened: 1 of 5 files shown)


examples/supermassive-todomvc/package.json
```diff
@@ -3,7 +3,7 @@
   "license": "MIT",
   "main": "index.js",
   "private": true,
-  "version": "1.0.1-0",
+  "version": "1.0.1-1",
```
pavelglac, line 6: I have manually released the new version in order to update license info. Since the original version did not have license. But this package should be private and there is not reason to ever publish this example.

alloy (approved):

PR #548, opened 2025-04-10
feat(apollo-forest-run): support custom logger and rudimentary telemetry

vladar:
# Logging and telemetry-related changes for ForestRun

## Custom logger

Replace all `console.log` statements with custom logger. Logger instance can be injected via cache config. The default is equivalent to:

```ts
const cache = new ForestRun({
  logger: console
});
```

Passing `logger: undefined` will disable logging (alternatively custom `log`, `debug`, `error`, `warn` methods could be used for fine-grain control over logging).

## Events for telemetry

Add ability to send structured data to telemetry, when corresponding notification handler is set:

```ts
const cache = new ForestRun({
  notify: (event) => {
    // event contains structured data to be sent to telemetry
  }
});
```

All event types are placed under `telemetry/types.ts`

### DIFF
--- a/packages/apollo-forest-run/src/cache/env.ts
+++ b/packages/apollo-forest-run/src/cache/env.ts
@@ -11,6 +11,7 @@ import {
 } from "../descriptor/types";
 import { getMergePolicyFn, getReadPolicyFn } from "./policies";
 import { SourceObject } from "../values/types";
+import { logger } from "../jsutils/logger";
 
 export function createCacheEnvironment(config?: CacheConfig): CacheEnv {
   const possibleTypes = config?.possibleTypes;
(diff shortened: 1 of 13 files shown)


packages/apollo-forest-run/src/cache/env.ts
```diff
@@ -37,6 +38,8 @@ export function createCacheEnvironment(config?: CacheConfig): CacheEnv {
     autoEvict: config?.autoEvict ?? true,
     nonEvictableQueries: config?.nonEvictableQueries ?? new Set(),
     maxOperationCount: config?.maxOperationCount ?? 1000,
+    logger: config && "logger" in config ? config.logger : logger,
```
pavelglac, line 41: question: based on this the logger should never be undefined. Why in functions props the logger is possibly undefined?

PR #549, opened 2025-04-10
feat(apollo-forest-run): cache extract

vladar:
Support `extract()` method from ApolloCache interface in ForestRun. Now custom changes we did for DevTools are not necessary, devtools can rely on common interface.


vejrj (approved):

hanys7 (commented):
ok

PR #552, opened 2025-04-22
fix(apollo-forest-run): later chunks should have precedence

vladar:
This PR addresses a difference in behavior between ForestRun and InMemoryCache in a peculiar scenario when written result contains conflicting data. 

In the following example InMemoryCache will pick value `foo2` over `foo1` when writing to cache:

```js
  const badResult = {
    foo1: { __typename: "Foo", id: "foo", value: "foo1" },
    foo2: { __typename: "Foo", id: "foo", value: "foo2" },
  };
```

Before this PR, ForestRun would pick `foo1` over `foo2`. With this PR it behaves the same way as InMemoryCache.


vladar:
This is an incomplete solution. For now we must accept this difference in behavior and assume proper graphql contract (i.e. no two nodes with different field values).

PR #555, opened 2025-05-15
WIP: feat(forest-run): logging expensive updates

pavelglac:


scripts/config/jest.config.js
```diff
@@ -1,7 +1,6 @@
 module.exports = {
   preset: "ts-jest",
   rootDir: process.cwd(),
-  roots: ["<rootDir>/src"],
```
vladar, line 4: I am also struggling with this one :) I think we need to refactor the config for CI vs local development (either with env variables or just have two configs).

Not for this PR of course. Also don't forget to remove this in the end :)

PR #557, opened 2025-05-26
feat(forest-run): logging update metadata

pavelglac:


packages/apollo-forest-run/src/forest/types.ts
```diff
@@ -44,6 +44,9 @@ export type IndexedTree = {
 
   // ApolloCompat
   danglingReferences?: Set<NodeKey>;
+
+  // Metadata about chunk updates, used for telemetry and debugging
+  chunksMetadata: UpdateChunkMetadata[];
```
vladar, line 49: After some consideration I now think we should not put update details to IndexedTree. Let's make `updateTree` return `UpdateState` and include update stats + indexedTree in this state object. It is already created today, so the overhead should be minimal.

packages/apollo-forest-run/src/forest/types.ts
```diff
@@ -57,6 +60,18 @@ export type IndexedForest = {
 export type Source = Readonly<SourceObject | SourceCompositeList>;
 export type Draft = SourceObject | SourceCompositeList;
 
+export type UpdateChunkMetadata = {
```
vladar, line 63: Let's stick with more specific "Stats" suffix here vs "Metadata"

```suggestion
export type UpdateChunkStats = {
```

packages/apollo-forest-run/src/forest/types.ts
```diff
@@ -57,6 +60,18 @@ export type IndexedForest = {
...
+  type: string;
+  depth: number;
+  siblings: number;
+  dirtyFields: Set<string>;
+};
+
+export type UpdateForestMetadata = {
```
vladar, line 70: Ditto
```suggestion
export type UpdateForestStats = {
```

packages/apollo-forest-run/src/forest/types.ts
```diff
@@ -57,6 +60,18 @@ export type IndexedForest = {
...
+  totalUpdated: number;
```
vladar, line 71: This will likely be the same as `operationCosts.size`? So maybe just return the map with update stats for now?

packages/apollo-forest-run/src/forest/types.ts
```diff
@@ -57,6 +60,18 @@ export type IndexedForest = {
...
+  operationCosts: Map<OperationDescriptor, UpdateChunkMetadata[]>;
```
vladar, line 72: WDYT if we rename it to `updateStats`? Or if we return the map itself, we probably don't need this whole type.

packages/apollo-forest-run/src/telemetry/updateStats/updateLogger.ts
```diff
@@ -0,0 +1,155 @@
...
+export function makeCopyStats(): CopyStats {
+  return {
+    arraysCopied: 0,
+    arrayItemsCopied: 0,
+    objectFieldsCopied: 0,
+    objectsCopied: 0,
+  };
+}
+
+export class UpdateLogger {
```
vladar, line 41: For code style consistency, I would prefer to not use classes here. But this is not urgent, so up to you for this iteration.

vladar (approved):
Nice work! Added a couple of suggestions inline, nothing blocking though

packages/apollo-forest-run/src/telemetry/updateStats/updateLogger.ts
pavelglac: I get inspirated from already existing stats collectors: https://github.com/microsoft/graphitation/blob/user/pavelglac/update-metadata/packages/apollo-forest-run/src/cache/stats.ts. I would prefer to skip the refactor in this iteration

PR #559, opened 2025-05-30
[ts-codegen] Implement context types with namespaces defined in the input file

vejrj:
We want to enable people to define the resolver context subtypes  with namespaces.

Expected input: 
--context-sub-type-metadata-file
```
{
      baseContextTypePath: "@package/default-context",
      baseContextTypeName: "DefaultContextType",
      contextTypes: {
        managers: {
          user: {
            importNamespaceName: "UserStateMachineType",
            importPath: "@package/user-state-machine",
          },
          whatever: {
            importPath: "@package/whatever-state-machine",
          },
}
```

### DIFF
--- a/packages/cli/src/supermassive.ts
+++ b/packages/cli/src/supermassive.ts
@@ -17,14 +17,12 @@ type GenerateInterfacesOptions = {
   legacy?: boolean;
   legacyModels?: boolean;
   useStringUnionsInsteadOfEnums?: boolean;
-  contextSubTypeMetadataFile?: string;
+  contextTypeExtensionsFile?: string;
   enumMigrationJsonFile?: string;
   enumMigrationExceptionsJsonFile?: string;
   generateOnlyEnums?: boolean;
   generateResolverMap?: boolean;
   mandatoryResolverTypes?: boolean;
-  defaultContextSubTypePath?: string;
-  defaultContextSubTypeName?: string;
   scope?: string;
 };
 
@@ -56,7 +54,7 @@ export function supermassive(): Command {
     )
     .option("-cn, --context-type-name [contextTypeName]", "Context type name")
     .option(
-      "-dcp, --default-context-sub-type-path [defaultContextSubTypePath]",
+      "-dcp, --default-context-sub-type-path [baseContextSubTypePath]",
       "From where the default context type will be exported",
     )
     .option(
@@ -81,8 +79,8 @@ export function supermassive(): Command {
       "File containing array of enum names, which should remain typescript enums",
     )
     .option(
-      "--context-sub-type-metadata-file [contextSubTypeMetadataFile]",
-      "Subtype metadata file",
+      "--context-sub-type-metadata-file [contextTypeExtensionsFile]",
+      "Describes context types and their import paths. Used to generate resolver context type extensions. The file must be defined in the following format: { baseContextSubTypePath?: string, baseContextSubTypeName?: string, contextSubTypes: { [namespace: string]: { [type: string]: { importNamespaceName: string, importPath: string }}}",
     )
     .option(
       "--generate-resolver-map",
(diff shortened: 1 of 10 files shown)


packages/ts-codegen/src/__tests__/context.test.ts
```diff
@@ -2,10 +2,81 @@ import ts from "typescript";
...
+import { SubTypeNamespace } from "../codegen";
+import { OutputMetadata } from "../context/utilities";
 
 describe(generateTS, () => {
   describe("Tests basic syntax GraphQL syntax", () => {
+    const contextSubTypeMetadata = {
+      managers: {
+        user: {
+          name: "user",
+          importTypeName: 'UserStateMachineType["user"]',
```
user1736, line 14: what's the point of `importTypeName`? from the tests it seems like `UserStateMachineType` part is generated according some convention from `importPath` and `"user"` part always matches `name`. What's the point to make user type it?

packages/ts-codegen/src/__tests__/context.test.ts
```diff
@@ -2,10 +2,81 @@ import ts from "typescript";
...
-import { ContextMap } from "../context";
+import { SubTypeNamespace } from "../codegen";
+import { OutputMetadata } from "../context/utilities";
 
 describe(generateTS, () => {
   describe("Tests basic syntax GraphQL syntax", () => {
+    const contextSubTypeMetadata = {
+      managers: {
+        user: {
+          name: "user",
```
user1736, line 13: why this config field is needed? from what I see in the tests in always matches the `key` one level above it.

packages/ts-codegen/src/__tests__/context.test.ts
```diff
@@ -2,10 +2,81 @@ import ts from "typescript";
...
+          importTypeName: 'UserStateMachineType["user"]',
+          importPath: "@package/user-state-machine",
```
user1736, line 15: is it actually a `path`? can I use relative paths here?

packages/ts-codegen/src/__tests__/context.test.ts
```diff
@@ -47,29 +118,88 @@ describe(generateTS, () => {
             }
           `,
           {
-            contextSubTypeNameTemplate: "I${resourceName}StateMachineContext",
-            contextSubTypePathTemplate:
-              "@msteams/core-cdl-sync-${resourceName}",
+            contextSubTypeMetadata: contextSubTypeMetadata,
+            defaultContextSubTypePath: "@package/default-context",
+            defaultContextSubTypeName: "DefaultContextType",
```
user1736, line 123: given how it's used I think it's due to rename this to `baseContext*` or something with the same meaning

e.g. `baseContextTypePath` and `baseContextTypeName`

packages/ts-codegen/src/__tests__/context.test.ts
```diff
@@ -47,29 +118,88 @@ describe(generateTS, () => {
             }
           `,
           {
-            contextSubTypeNameTemplate: "I${resourceName}StateMachineContext",
-            contextSubTypePathTemplate:
-              "@msteams/core-cdl-sync-${resourceName}",
+            contextSubTypeMetadata: contextSubTypeMetadata,
```
user1736, line 121: wdyt about `contextTypeExtension` (or `contextTYpeExtensions`)?

packages/cli/src/supermassive.ts
```diff
@@ -88,6 +80,10 @@ export function supermassive(): Command {
       "--enum-migration-exceptions-json-file [enumMigrationExceptionsJsonFile]",
       "File containing array of enum names, which should remain typescript enums",
     )
+    .option(
+      "--context-sub-type-metadata-file [contextSubTypeMetadataFile]",
+      "Subtype metadata file",
```
user1736, line 85: would be great to have a more detailed comment here, possibly with file format example

user1736 (changes requested):
I didn't look much into implementation yet, because config format left me quite puzzled.

packages/ts-codegen/src/__tests__/context.test.ts
vejrj: yes, it can be either package or relative path

packages/ts-codegen/src/__tests__/context.test.ts
vejrj: that's format we agreed on with CDL

PR #675, opened 2026-06-16
shell-quote update

celiac747:


sergey-stoyan:
Do we need to add change file as well?

celiac747:
> Do we need to add change file as well?

beachball doesn't detect this type of change.. I don't know how is set up but in this case beachball says that there are no changes

sergey-stoyan (approved):

PR #676, opened 2026-06-17
chore(yarn): ensure network isolation for the build and tests

pavelglac:
## Why
Upgrade to yarn4 and ensure network isolation for the install, build and tests. For now there is not network isolation for the npm release (for the internal release there is). This will be addressed by bumping beachball and change the release pipeline according to https://github.com/microsoft/beachball/pull/1281


.azure-devops/graphitation-release.yml
```diff
@@ -46,16 +50,37 @@ extends:
...
               - script: |
                   yarn ci
                 displayName: build and test [test]
+                env:
+                  NPM_CONFIG_USERCONFIG: $(Agent.TempDirectory)/npm-mirror.npmrc
               - script: |
                   git config user.email "gql-svc@microsoft.com"
                   git config user.name "Graphitation Service Account"
-                  git fetch --depth=2
+                  git fetch origin main
```
user1736, line 83: you sure you don't want to keep depth for faster checkout? fetching all history looks like overkill here

pavelglac: Without it I was getting: fatal: shallow file has changed since we read it. For performance I would not be bothered.

image

.azure-devops/graphitation-release.yml
```diff
@@ -22,6 +24,8 @@ extends:
   template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
 
   parameters:
+    # settings:
+    #   networkIsolationPolicy: Permissive,CFSClean
```
user1736, line 28: should it be commeneted?

pavelglac: Yes, currently our pipeline cannot run in network isolation because of npmjs and github enpoints during the release.

celiac747 (approved):

PR #687, opened 2026-07-09
[WIP] add fragment loader support to variables/args processing

user1736:
Enables schemaFragmentLoader usage during argument and variables processing. Before it would hard fail if required input types were missing in current schema fragment

### DIFF
--- a/packages/supermassive/src/values.ts
+++ b/packages/supermassive/src/values.ts
@@ -6,6 +6,7 @@ import {
   DirectiveNode,
   FieldNode,
   VariableDefinitionNode,
+  ArgumentNode,
 } from "graphql";
 import { inspect } from "./jsutils/inspect";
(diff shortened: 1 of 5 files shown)


packages/supermassive/src/values.ts
```diff
@@ -296,7 +479,18 @@ export function getDirectiveValues(
...
   if (directiveNode) {
-    return getArgumentValues(exeContext, directiveDef, directiveNode);
+    const argValues = getArgumentValues(
+      exeContext,
+      directiveDef,
+      directiveNode,
+    );
+
+    // FIXME: check if schemaLoader should load types for directives
+    if (isPromise(argValues)) {
```
user1736, line 489: FIXME: stubbed it out for now, to do a smoke test in TMP

PR #693, opened 2026-07-14
feat(release): beachball esrp release

pavelglac:


.azure-devops/graphitation-release.yml
```diff
@@ -8,8 +8,8 @@ variables:
     value: production,externalfacing
   - name: serviceTreeID
     value: ade7d667-42f5-485a-91a9-f1dc6482a9b0
-  - name: adoNpmFeedBaseUrl
```
pavelglac, line 11: Maybe I will try to figure out how to do canary release but now it is out of the scope.

.azure-devops/graphitation-release.yml
```diff
@@ -34,39 +36,37 @@ extends:
     stages:
       - stage: release
         variables:
-          # OPTIONAL: Set this varibale to 'true' to enable signing in a target stage.
-          # Remove if signing is not required.
-          Build.ESRP.CodeSign.Enabled: true
+          # ESRP publishing is handled by EsrpRelease@9.
+          Build.ESRP.CodeSign.Enabled: false
```
celiac747, line 40: Is this change intended? why we need to remove code signing?

pavelglac:
Test package will be removed once I will be able to publish package (currently I need to wait a few hours to use new beachball version).

.azure-devops/graphitation-release.yml
celiac747: Discussed offline

.azure-devops/graphitation-release.yml
```diff
@@ -77,40 +77,105 @@ extends:
...
+
               - task: npmAuthenticate@0
+                displayName: npm authenticate for private ADO feed
                 inputs:
                   workingFile: ".npmrc"
+
               - script: |
                   releaseBranch="origin/${BUILD_SOURCEBRANCH#refs/heads/}"
+                  yarn beachball publish -b "$releaseBranch" -t latest -y --registry $(adoNpmFeedBaseUrl) --no-push --keep-change-files
+                displayName: Release to the ADO npm feed
```
sergey-stoyan, line 132: It's confusing a bit: in job "Build" we now have steps for release, and then we have a separate "Release" job

package.json
```diff
@@ -26,7 +26,7 @@
   "devDependencies": {
     "@babel/core": "^7.14.0",
     "@babel/preset-typescript": "^7.14.0",
-    "beachball": "^2.47.1",
+    "beachball": "3.0.0-alpha.4",
```
sergey-stoyan, line 29: Why do we upgrade to this alpha version?

pavelglac: yes, we need to have `pack-to-path` option

.azure-devops/graphitation-release.yml
pavelglac: Agree. The logic was that is expesive to have it as separate job because you need to install dependencies and build packages. Now the job for releasing to the ADO takes 8 min.

Image

yarn.lock
```diff
@@ -7895,25 +7909,23 @@ __metadata:
...
-  version: 2.47.1
-  resolution: "beachball@npm:2.47.1"
+"beachball@npm:3.0.0-alpha.4":
+  version: 3.0.0-alpha.4
+  resolution: "beachball@npm:3.0.0-alpha.4"
   dependencies:
-    cosmiconfig: "npm:^8.3.6"
-    execa: "npm:^5.0.0"
-    fs-extra: "npm:^11.1.1"
-    lodash: "npm:^4.17.15"
```
celiac747, line 7905: This dependency is removed, is it expected?

celiac747: I think I remember from when I was upgrading lodash that it needed to upgrade beachball to don't pin it so I guess we should be fine

celiac747 (approved):

.azure-devops/graphitation-release.yml
sergey-stoyan: Can we have release process described in some "RELEASE.md" or something? Because it's really not obvious now what's going on

sergey-stoyan (approved):

yarn.lock
pavelglac: I have bumped the beachball: https://github.com/microsoft/graphitation/pull/693/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R29 so it is expected

PR #704, opened 2026-08-04
fix(apollo-forest-run): report the malformed payload behind a corrupted list instead of dereferencing a hole

vladar:
# Report the malformed payload behind a corrupted list instead of dereferencing a hole

Fixes the production `TypeError: Cannot read properties of undefined (reading 'value')` from
`reIndexList` (`src/forest/indexTree.ts`), reported against 0.24.5 for
`ComponentsChatQueriesMessageListQuery`.

**Scope: this only improves the error raised when the condition is hit.** It does not change what the
cache accepts, and does not stop the bad state from being created — both are follow ups below.

## The defect

A payload may contain the same node twice, and indexing aggregates the occurrences — sound only when
they agree. When two occurrences carry **different lengths for the same list field**,
`resolveListItemChunk` applies an index valid for the longer one to the shorter chunk. `itemChunks`
is a lazily populated sparse cache of `data`, so assigning past the end grows it; if the first out of
range index is strictly greater than `data.length`, a **hole** is left behind.

That hole sits in committed cache state and reads fine (reads go through `data`). It only surfaces
when a later write recycles the chunk, since `reIndexList` is the one place that walks every item
reference — which is why the operation in the stack trace is innocent.

## The change

`reIndexList` checks the reference before dereferencing it. Detection happens there, but the message
describes the write that *created* the hole: the tree being recycled still holds the offending
payload, so both occurrences are reconstructed from it.

```
Invariant violation: Detected malformed payload written to the cache: a "Thread" node occurs multiple times in a single write with a different number of items in the "messages" list.

  Operation:  query MessageList
  Node type:  Thread
  Node id:    same in both occurrences (not shown)
  Field:      messages

  Occurrence 1: 14 items at data.conversation.thread.messages
  Occurrence 2: 3 items at data.pinned.thread.messages
```

`Operation` is the write that produced the payload — recycling is always same-operation, so there is
no second one to name. `Node id` states what makes the divergence a conflict (both paths resolve to
the same entity) without printing the id.

Hot path cost is one `=== undefined` per list item; the message is built only on failure.

**Telemetry safety.** The `Invariant violation: ` prefix from `assert` is load bearing — the
consuming Apollo wrapper allowlists on `startsWith("Invariant")` and *replaces* non-matching messages
with a character count. Everything printed is schema level: type names, field names, data paths,
counts. Never the node key (it embeds the entity id) or argument values.

## Tests

`src/__tests__/regression.test.ts`, public cache API only.

| test | what it pins |
| --- | --- |
| `reports an unresolved list item instead of dereferencing it` | the production shape: one write punches the hole, the next recycles it; asserts the full message and that the entity id does not leak |
| `reports list indices when the repeated node is itself a list item` | repeated node is a list item with the divergent list hanging off it — occurrences render with their index (`data.inbox.messages.2.files`), and a single item reads `1 item` |
| `accepts a payload repeating a node with lists of equal length` | equal lengths still recycle cleanly — guards against over firing |

Both failing tests return the original error when the guard alone is reverted:

```
Received string: "Cannot read properties of undefined (reading 'value')"
```

Suite: 25 suites, 1101 passed, 12 skipped, 8 todo.

## Follow ups

Both prevent the bad state rather than report it, which is why neither is here.

- **Validate the payload at index time.** Would blame the right operation *and* catch the variant
  where out of range indices resolve in order — the chunk densifies, no hole is left, and this check
  sees nothing. Costs a traversal over repeated nodes and needs its own risk assessment.
  Marked `test.todo("rejects repeated nodes with divergent list lengths at index time")`.
- **Bounds check in `resolveListItemChunk`.** Stops the state being created at all, for every
  producer; a disabled assert for exactly this already sits a few lines above the write. It changes
  out of range semantics (fabricated absence becomes real absence), so it deserves its own PR.


vladar:
Superseded by #708 - same change, squashed onto current main. Closing this one because the review threads here made the diff hard to read; all feedback from them is incorporated in the new PR.

PR #724, opened 2026-09-01
Non mutating resolver merge

user1736:
This fixes an issue, where mutation of nested level objects (i.e. `resolvers.Query`, `resolvers.Mutation`) was causing resolvers leaking into modules they don't belong in TMP.

### DIFF
--- a/change/@graphitation-supermassive-21234229-4768-48a1-9465-f8a89f50dbf5.json
+++ b/change/@graphitation-supermassive-21234229-4768-48a1-9465-f8a89f50dbf5.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "feat: do not mutate nested objects on resolver merge",
+  "packageName": "@graphitation/supermassive",
+  "email": "dsamsonov@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


vladar (commented):
Thanks for tackling this — the cross-module resolver leaking has bitten us before, so I'm glad to see it addressed! I did dig into the new `mergeResolversObjMap` a bit and I think there are a few edge cases worth handling before we merge. I verified each of these locally against the repo's `graphql`.

### 1. Custom scalars get quietly downgraded

`ScalarTypeResolver = GraphQLScalarType` (`types.ts:18`), so a resolver map can legitimately hold a class instance. Copying it into a plain `{}` drops the prototype:

```
orig isScalarType: true
copy isScalarType: false
```

We could get away with it before this change because of direct usage of the package with custom scalars.

### 2. Perf: we now copy on every merge

Since this is meant to be the fast alternative to graphql-tools, it's a little unfortunate that we now allocate an object and shallow-copy every field for every type, even in the very common case where only one module contributes a given type. Copy-on-write would keep the fast path and happens to fix (1), (2) and (3) at the same time:

```ts
function mergeResolversObjMap(
  accumulator: Resolvers,
  resolvers: Resolvers,
  owned: Set<string>,
) {
  for (const [typeName, typeResolver] of Object.entries(resolvers)) {
    const existing = accumulator[typeName];
    if (existing === undefined) {
      if (typeResolver) accumulator[typeName] = typeResolver; // fast path, no copy
      continue;
    }
    if (!isObjectLike(existing) || !isObjectLike(typeResolver)) continue;
    if (!owned.has(typeName)) {
      // clone only when a second module actually contributes to this type
      accumulator[typeName] = Object.assign(
        Object.create(Object.getPrototypeOf(existing)),
        existing,
      );
      owned.add(typeName);
    }
    Object.assign(accumulator[typeName], typeResolver);
  }
}
```

A couple of notes on that sketch: `owned` needs to be created once per top-level `mergeResolvers` call and threaded through the array recursion, since entries the caller pre-seeded into the accumulator aren't ours to mutate. And `Object.create(Object.getPrototypeOf(existing))` is what keeps `instanceof GraphQLScalarType` intact.

vladar (approved):

PR #732, opened 2026-09-09
Fragment loader support for variables

user1736:
Enables schemaFragmentLoader support during variables processing. Before it would hard fail if required input types were missing in current schema fragment.

support for argument processing (`getArgumentValues`) is out of the scope for this PR.

### DIFF
--- a/packages/supermassive/src/executeWithoutSchema.ts
+++ b/packages/supermassive/src/executeWithoutSchema.ts
@@ -48,6 +48,7 @@ import {
   getArgumentValues,
   getVariableValues,
   getDirectiveValues,
+  getMissingVariableTypes,
 } from "./values";
 import type { ExecutionHooks } from "./hooks/types";
 import { arraysAreEqual } from "./utilities/array";
@@ -115,7 +116,7 @@ export interface ExecutionContext {
   contextValue: unknown;
   buildContextValue?: (contextValue?: unknown) => unknown;
   operation: OperationDefinitionNode;
-  variableValues: { [variable: string]: unknown };
+  variableValues: VariableValues;
   fieldResolver: FunctionFieldResolver<unknown, unknown>;
   typeResolver: TypeResolver<unknown, unknown>;
   subscribeFieldResolver: FunctionFieldResolver<unknown, unknown>;
@@ -125,6 +126,8 @@ export interface ExecutionContext {
   enablePerEventContext: boolean;
   enableEarlyExecution: boolean;
 }
+
+type VariableValues = { [variable: string]: unknown };
 
 /**
  * Implements the "Executing requests" section of the GraphQL specification.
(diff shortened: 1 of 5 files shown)


packages/supermassive/src/executeWithoutSchema.ts
```diff
@@ -141,14 +147,31 @@ export function executeWithoutSchema(
...
-
+  const baseContext = buildExecutionContext(args);
   // Return early errors if execution context failed.
-  if (!("schemaFragment" in exeContext)) {
-    return { errors: exeContext };
-  } else {
-    return executeOperationWithBeforeHook(exeContext);
+  if (!("schemaFragment" in baseContext)) {
+    return { errors: baseContext };
+  }
```
vladar, line 151: Instead of doing `baseContext` thing - can we extract validation that is performed inside `buildExecutionContext` to a separate function, run it first thing, return errors if any. 

If no errors - do coercion with just args vs. the context. Then create the actual context once (and pass coercedValues into it).

packages/supermassive/src/values.ts
```diff
@@ -45,7 +48,7 @@ type CoercedVariableValues =
  *
  * @internal
  */
-export function getVariableValues(
```
vladar, line 48: Can we not rename it? It will make maintenance vs graphql-js a bit more confusing because it also has `coerceVariableValues` and now it will be pointing to a different function in our codebase.

packages/supermassive/src/executeWithoutSchema.ts
```diff
@@ -170,6 +193,48 @@ export function assertValidExecutionArguments(
...
 
+function coerceVariableValuesWithSchemaLoader(
+  rawVariableValues: Maybe<VariableValues>,
+  baseContext: BaseExecutionContext,
+): PromiseOrValue<Array<GraphQLError> | VariableValues> {
+  const { operation, schemaFragment } = baseContext;
+  // istanbul ignore next (See: 'https://github.com/graphql/graphql-js/issues/2203')
+  const variableDefinitions = operation.variableDefinitions ?? [];
+
+  const missingTypes = getMissingVariableTypes(
```
vladar, line 198: I just realized that `getVariableValues` (`coerceVariableValues` in this PR) has onError callback. Which makes me wonder - do we even need `getMissingVariableTypes` - we can just include missing type name in the error it triggers, no?

packages/supermassive/src/executeWithoutSchema.ts
```diff
@@ -665,32 +717,6 @@ function executeField(
   });
 }
 
-function requestSchemaFragment(
```
vladar, line 668: Nit: can we not extract it into utilities. This is not a utility, but a key part of execution in our case.

user1736: Fixed, but it'd be nice to eventually do some splitting there. The file is already 3k+ lines.

packages/supermassive/src/values.ts
user1736: okay, if it's easier to navigate.

packages/supermassive/src/executeWithoutSchema.ts
user1736: I tried to do that, but couldn't find a nice way to do it. `requestSchemaFragment` requires access to `exeContext` object , so it can set `mergedFragment` and `mergedContextValue`.

Extracting operation validation from `buildExecutionContext` it also turned out to be wack, because then you'd need to traverse definitions twice: once inside validation and the other one to collect fragments.

packages/supermassive/src/executeWithoutSchema.ts
user1736: I tried to adapt it, but it'd require function redesign to make it work. Right now it doesn't support partial coercion. So keeping this part as is, since we want to keep `getVariableValues` closer to its graphql-js counterpart.

vladar (approved):

